PR #3, opened 2019-10-20
KK-21 | Model changes and initial API

tuomas777:
Model changes:
* Removed child SSN and everything related to that
* Added a new model for guardians (because we need to store
guardian specific data, and not every user will be a guardian)

API:
* Implemented initial mutation for submitting child data
(+ guardian data and relationship type related to the child)

Tests will be implemented separately later.


children/schema.py
```diff
@@ -0,0 +1,95 @@
...
+    guardian = graphene.Field(GuardianType)
+
+    @classmethod
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO we'll probably need to check somehow if the child already exists
+        child = Child.objects.create(**kwargs["child"])
+
+        # TODO change this to logged in user once we have authentication in place
+        user_obj, _ = User.objects.get_or_create(username="dummyman777")
```
igordavydsson: I guess this is good for the time being, until the logging in is implemented. Might as well do just `.create()` maybe, since you create the `Child` on each mutation anyway.

tuomas777: True, that could actually be better.

PR #5, opened 2019-10-31
KK-27 | Add tests for SubmitChild mutation

tuomas777:


children/tests/test_api.py
```diff
@@ -0,0 +1,82 @@
...
+                "child": CHILD_DATA,
+                "relationship": {"type": "PARENT"},
+                "guardian": GUARDIAN_DATA,
+            }
+        }
+    }
+
+    child = Child.objects.get(**to_snake_dict(CHILD_DATA))
+    guardian = Guardian.objects.get(**to_snake_dict(GUARDIAN_DATA))
+    Relationship.objects.get(type=Relationship.PARENT, child=child, guardian=guardian)
```
igordavydsson: maybe slightly more verbose would be more descriptive:
```python
assert Relationship.objects.filter(type=Relationship.PARENT, child=child, guardian=guardian).exists()
```

tuomas777: True, but then it is not consistent with the two other checks, child and guardian. And consistency is my middle name. I think I'll leave it that way, but add a comment to make the whole thing clear :)

igordavydsson: alright 👍

PR #6, opened 2019-10-31
KK-53 | Refactor Docker setup

tuomas777:
Pretty much copypasted from https://github.com/City-of-Helsinki/open-city-profile/tree/87104d7c28cfc1e385e90252bbf0fd60dd6d6812


docker-compose.yml
```diff
@@ -1,35 +1,39 @@
-version: '3'
+version: '3.7'
 services:
     postgres:
-        build: ./docker/postgres/
+        image: helsinkitest/postgis:9.6-2.5-alpine
```
igordavydsson: should we already use `10-2.5` maybe?

badrange: Do we need postgis features in this project? If there is no plain postgres image, I have no problems using a postgis image, on the other hand upgrades between postgres versions are simpler if we don't have the postgis extensions..

tuomas777: We don't need postgis yet, but we might need it in the future when we add places for events and stuff. That said, I would have used plain postgres, but there is no HKI image for that, and I thought we should use only our own images. But I learned that is not the case, so changing to to official postgres image.

And using v10 is better, changing to that. We should actually probably go with 11 as soon as we can be sure that it doesn't cause problems anywhere (it used to cause problems with Travis at least).

tuomas777:
Changed python image to HKI one and postgis to official postgres v10. Did also some tweaking to `Dockerfile` suggested by @frwickst

PR #9, opened 2019-11-12
KK-63 | Add children and guardians GraphQL queries

tuomas777:
Normal users can view only their own children and guardian info while admin users have no restrictions. Currently being an admin is determined by is_staff attribute on User model.

[snapshottest](https://github.com/syrusakbary/snapshottest) is added for tests.


children/schema.py
```diff
@@ -22,11 +34,10 @@ class ChildInput(graphene.InputObjectType):
     birthdate = graphene.Date(required=True)
 
 
-# this class cannot be named just RelationshipType because of graphene-django
-# https://github.com/graphql-python/graphene-django/issues/185
-class RelationshipObjectType(DjangoObjectType):
+class RelationshipNode(DjangoObjectType):
```
igordavydsson: yes, very nice! we should indeed call them `...Node` if they implement `Node interface` 👍 should be part of the GraphQL guidelines.

children/schema.py
```diff
@@ -96,3 +102,12 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         return SubmitChildMutation(
             child=child, relationship=relationship, guardian=guardian
         )
+
+
+class Query:
+    children = relay.ConnectionField(ChildConnection)
```
igordavydsson: you can also do:
```
from graphene_django import DjangoConnectionField

...

class Query:
    children = DjangoConnectionField(ChildNode)

    ...

```
this way you don't even hav to declare `ChildConnection`, as graphene will generate it for you.

users/schema.py
```diff
@@ -0,0 +1,32 @@
...
+        return queryset.filter_for_user(info.context.user).order_by("last_name")
+
+
+class GuardianConnection(relay.Connection):
+    class Meta:
+        node = GuardianNode
+
+
+class Query:
+    guardians = relay.ConnectionField(GuardianConnection)
```
igordavydsson: same here

children/schema.py
igordavydsson: having an explicit `Connection` class is called for, if you need to add some of your own sugar to it, like Pekka did in [profile](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/schema.py#L59)

users/tests/snapshots/snap_test_api.py
```diff
@@ -0,0 +1,91 @@
...
+snapshots = Snapshot()
+
+snapshots["test_guardians_query_normal_user 1"] = {
+    "data": {
+        "guardians": {
+            "edges": [
+                {
+                    "node": {
+                        "email": "angelahawkins@gmail.com",
+                        "firstName": "Jessica",
```
igordavydsson: I don't get it 😅 why would it match every time, if it just comes from `...Factory()`. doesn't factory generate new names and such on each run?

igordavydsson: like, if I pull this branch and run `pytest` I would get data that would not match with the snapshot? or how does it work 🙈

children/schema.py
tuomas777: That is actually a bit odd IMO. For normal `ConnectionField` you need an explicit `Connection` now (they seem to have changed it to work like that), but for `DjangoConnectionField` you don't.

But yeah I'll change those, good point.

users/tests/snapshots/snap_test_api.py
tuomas777: The factories will generate the same values everywhere everytime (or at least they should :) ) because the random seed for those is set to a fixed value [here](https://github.com/City-of-Helsinki/kukkuu/pull/9/files#diff-5c3f086ee36c1e0e4e81d69639021286R13).

children/schema.py
igordavydsson: yeah, it was a change in Graphene 2.0. don't know why it only affect regular ConnectionFields.

users/tests/snapshots/snap_test_api.py
igordavydsson: aaaahh, now it all makes sense. 👍 thanks for clarigying!

children/schema.py
tuomas777: Changed.

users/tests/snapshots/snap_test_api.py
tuomas777: Actually, not sure how robust the thing is after all, as changing the connectionfields seems to have affected the snapshots :D 

I continue to evaluate the snapshot thing in this project and let's see how well it works for us. It sure seems handy if it works.

PR #11, opened 2019-11-22
KK-110 & KK-111 | Add signup notification

tuomas777:


children/notifications.py
```diff
@@ -0,0 +1,22 @@
+from django.utils.translation import ugettext_lazy as _
+from django_ilmoitin.dummy_context import dummy_context
+from django_ilmoitin.registry import notifications
+
+from users.factories import GuardianFactory
+
+from .factories import ChildWithGuardianFactory
+
+
+class NotificationType:
```
igordavydsson: any particular reason for a basic object instead of `Enum`? works either way, just wondering

tuomas777: If I'm not mistaken, for enums django-ilmoitin currently requires you to use `NotificationType.PIUPAU.value` everywhere instead of just `NotificationType.PIUPAU`.  That is not really nice and it makes shooting yourself in the foot easy if you mix the value and the enum itself somewhere (been there with haravajärjestelmä :) ). 

So as kukkuu does not even use enum fields, I figured it would be simpler to just use a basic object. When we get proper support for enums implemented in django-ilmoitin, I will probably migrate the type to an enum :)

igordavydsson: you are right about the enum support not being the best. 😅 alright then, object makes a lot of sense.

PR #12, opened 2019-11-26
KK-116 | Children mutation and model tweaks

tuomas777:


users/models.py
```diff
@@ -30,6 +30,7 @@ class Guardian(UUIDPrimaryKeyModel, TimestampedModel):
     phone_number = models.CharField(
         verbose_name=_("phone number"), max_length=64, blank=True
     )
+    language = models.CharField(verbose_name=_("language"), max_length=10, blank=True)
```
igordavydsson: should the validation be enforced already here by passing the `options=settings.LANGUAGES` maybe?

tuomas777: Yeah that would be nice, however it would also mean that new migrations are created whenever the languages are changed, which I don't like.

I guess I don't really need the model validation now, but what I should have had from the beginning is a proper select for the field in admin UI, which I added now :)

tuomas777:
Did some tweaking to the tweaks:

* changed language to be a required field
* added email field back for guardians in API queries, it is populated from the corresponding user's email

users/models.py
igordavydsson: makes sense 👍

PR #13, opened 2019-11-27
KK-97 | Add child update and delete mutations

tuomas777:


children/schema.py
```diff
@@ -125,10 +134,100 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+    postal_code = graphene.String()
+    relationship = graphene.Field(RelationshipNode)
+
+    @classmethod
+    @login_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **child_data):
+        user = info.context.user
+        child_global_id = child_data.pop("id")
+        child = Child.objects.user_can_view(user).get(
```
igordavydsson: will `.get` here behave ok and always return a child? can it potentially raise `DoesNotExist`, if say `user_can_view(user)` filters it out 🤔

children/schema.py
```diff
@@ -125,10 +134,100 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+class UpdateChildMutation(graphene.relay.ClientIDMutation):
+    class Input:
+        id = graphene.ID(required=True)
+        first_name = graphene.String()
+        last_name = graphene.String()
+        birthdate = graphene.Date()
+        postal_code = graphene.String()
+        relationship = RelationshipInput()
+
+    id = graphene.ID()
```
igordavydsson: isn't it easier to just return `ChildNode`?

children/schema.py
tuomas777: Yeah it can raise `DoesNotExist` but I figured that is ok for now, as it will result in an error response from the API. In the future we will probably need a way to send just some of those exceptions to Sentry, so this will need some love then.

children/schema.py
tuomas777: Yes that would be nicer in many ways, but it would also mean that the response fields need to be nested inside another field, which I don't really like in this particular situation.

tuomas777: And after some researching and rethinking, I think your way is superior, so I will switch to that :)

PR #16, opened 2019-12-04
KK-131 | Add child mutation

tuomas777:


common/tests/conftest.py
```diff
@@ -33,6 +33,13 @@ def staff_api_client():
     return _create_api_client_with_user(UserFactory(is_staff=True))
 
 
+@pytest.fixture
+def guardian_api_client():
+    return _create_api_client_with_user(
+        UserFactory(is_staff=True, guardian=GuardianFactory())
```
igordavydsson: does it need `is_stuff=True`?

children/tests/test_api.py
```diff
@@ -178,6 +188,60 @@ def test_children_query_staff_user(snapshot, staff_api_client):
...
+    )
+    assert_relationship_matches_data(
+        relationship, ADD_CHILD_VARIABLES["input"]["relationship"]
+    )
+
+
+@pytest.mark.django_db
+def test_add_child_mutation_birthdate_required(snapshot, user_api_client):
+    variables = deepcopy(UPDATE_CHILD_VARIABLES)
+    variables["input"].pop("postalCode")
```
igordavydsson: 1. test name kinda suggests that you wanted to pop `birthdate` 😅 
2. not sure why poping `postalCode` fails the mutation though, since it's non-required 🤔

common/tests/conftest.py
tuomas777: No, and should not have it either, good catch 👍

children/tests/test_api.py
tuomas777: Yeah, this test was actually flawed in a couple of ways, good thing you spotted it :)

This actually showed that just checking for "errors" to exists in the response isn't a very optimal way in this kind of tests because of false positives, so gotta figure out a better way for these in the future (but not in this PR :) )

tuomas777:
Fixed the mentioned issues.

children/tests/test_api.py
igordavydsson: heh, alright!

PR #19, opened 2019-12-11
KK-144 | Validations and fixes

tuomas777:
The PR contains a couple of added validations and in addition minor fixes to permission handling and tweaks to tests.

### DIFF
--- a/children/schema.py
+++ b/children/schema.py
@@ -1,5 +1,6 @@
 import graphene
 from django.contrib.auth import get_user_model
+from django.core.exceptions import ValidationError
 from django.db import transaction
 from django_ilmoitin.utils import send_notification
 from graphene import relay
@@ -13,7 +14,7 @@ from children.notifications import NotificationType
 from users.models import Guardian
 from users.schema import GuardianNode, LanguageEnum
 
-from .models import Child, Relationship
+from .models import Child, postal_code_validator, Relationship
 
 User = get_user_model()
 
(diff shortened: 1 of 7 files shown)


children/schema.py
```diff
@@ -89,8 +92,11 @@ class Input:
     @login_required
     @transaction.atomic
     def mutate_and_get_payload(cls, root, info, **kwargs):
-        user = info.context.user
+        children_data = kwargs["children"]
+        if not children_data:
+            raise GraphQLError("At least one child is required.")
```
igordavydsson: isn't it ensure by the changes you made to the inout args?

igordavydsson: maybe [like this](https://docs.graphene-python.org/en/latest/types/list-and-nonnull/#nonnull-lists)?

tuomas777: Yeah, AFAIK there is no way in GraphQL schema to force a list to have at least one item.

But your suggestion of forbidding nulls is also needed, we don't want to allow `[null]`, so added that :)

igordavydsson: hm, wouldn't the way you did it now enforce `a list to have at least one item`?
you get right now the following schema: `children: [ChildInput!]!`

igordavydsson: I think this bit is now redundant:
```
children_data = kwargs["children"]
if not children_data:
    raise GraphQLError("At least one child is required.")
```

tuomas777: Yes I get exactly that schema now, but it still allows `children: []` so that check is needed.

igordavydsson: is this a bug in graphene? 😅 man, well, we can keep the check I guess.

tuomas777: It is actually how it goes in GraphQL, see [this](https://stackoverflow.com/questions/46770501/graphql-non-nullable-array-list/46771890#46771890).

igordavydsson: oh, I see! thanks for the link. I had a different understanding of those `!` marks. 🙂 this PR is then 💯 ⚡️

PR #20, opened 2019-12-12
KK-154 | Send GraphQL errors to sentry

tuomas777:
Added logic that sends all exceptions to Sentry, with the exception (hehehe) of:
* Bad Request type exceptions
* custom `KukkuuGraphQLError`s, which are meant for returning errors from the API which we don't want to Sentry

The PR contains also one unrelated bug fix.

### DIFF
--- a/kukkuu/views.py
+++ b/kukkuu/views.py
@@ -1,0 +1,20 @@
+import sentry_sdk
+from graphene_django.views import GraphQLView
+
+
+class SentryGraphQLView(GraphQLView):
+    def execute_graphql_request(self, request, data, query, *args, **kwargs):
+        """Extract any exceptions and send some of them to Sentry"""
+        result = super().execute_graphql_request(request, data, query, *args, **kwargs)
+        # If 'invalid' is set, it's a bad request
+        if result and result.errors and not result.invalid:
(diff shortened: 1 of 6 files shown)


kukkuu/views.py
```diff
@@ -1,14 +1,22 @@
...
     def execute_graphql_request(self, request, data, query, *args, **kwargs):
         """Extract any exceptions and send some of them to Sentry"""
         result = super().execute_graphql_request(request, data, query, *args, **kwargs)
         # If 'invalid' is set, it's a bad request
         if result and result.errors and not result.invalid:
-            self._capture_sentry_exceptions(result.errors, query)
+            errors = [
+                e
+                for e in result.errors
+                if not isinstance(e.original_error, KukkuuGraphQLError)
```
pjlampil: I think you can't be sure that `error` always has the `original_error` attribute. Can't remember which case was it but I previously implemented like this:
```
        for error in errors:
            try:
                capture(error.original_error)
            except AttributeError:
                capture(error)
```
And I remember ending up catching AttributeError in some cases. But if it's only the cases where `result.invalid` I guess this should do it.

tuomas777: Good point, added a check for that.

PR #25, opened 2020-01-10
KK-169 | Add event, orrcurrence, venue models

quyenlq:


events/factories.py
```diff
@@ -0,0 +1,19 @@
...
+class EventFactory(factory.django.DjangoModelFactory):
+    name = factory.Faker("text", max_nb_chars=64)
+    duration = factory.Faker("random_int", max=1000)
+
+    class Meta:
+        model = Event
+
+
+class OccurrenceFactory(factory.django.DjangoModelFactory):
+    time = factory.Faker("date_time", tzinfo=tzoffset("UTC+2", 0))
```
tuomas777: I think it would be better if the timezone was the one actually in used Finland, so that DST will be taken into account. IIRC it can be set with something like `tzinfo=pytz.timezone("Europe/Helsinki")`

events/models.py
```diff
@@ -0,0 +1,30 @@
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from common.models import TimestampedModel, UUIDPrimaryKeyModel
+
+
+class Event(UUIDPrimaryKeyModel, TimestampedModel):
+    name = models.CharField(verbose_name=_("name"), max_length=64)
```
tuomas777: As we don't know what kind of names they want to give to the events, probably better to be safe than sorry and add longer `max_length` to start with, how about 255? :)

venues/models.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from common.models import TimestampedModel, UUIDPrimaryKeyModel
+
+
+class Venue(UUIDPrimaryKeyModel, TimestampedModel):
+    name = models.CharField(verbose_name=_("name"), max_length=64)
```
tuomas777: Maybe a longer name here also.

venues/models.py
```diff
@@ -0,0 +1,17 @@
...
+    description = models.TextField(verbose_name=_("description"), blank=True)
+    seat_count = models.IntegerField(verbose_name=_("seat count"))
```
tuomas777: This could be a `PositiveIntegerField`.

events/models.py
```diff
@@ -0,0 +1,30 @@
...
+    short_description = models.TextField(
+        verbose_name=_("short description"), blank=True,
+    )
+    description = models.TextField(verbose_name=_("description"), blank=True)
+    duration = models.IntegerField(verbose_name=_("duration"))
```
tuomas777: `Event` should have a `ForeignKey` to `Venue`, right? :)

quyenlq: According to the diagram, I believe event connects to an occurrence, which then connects to a venue, so basically an event can has multiple occurrences at different venues. Correct me if I'm wrong :D

venues/models.py
quyenlq: 👍 +1

events/models.py
quyenlq: Agree, 64 seems a bit short here

quyenlq:
> Added some comments, in general looking very good.
> 
> One thing about IDs. I don't think there is a reason why events, occurrences and venues couldn't have just an integer ID. One practical problem with UUIDs is that they make urls in UI look horrid, especially when using Graphene and Relay as we do in this project (that makes the IDs in the API longer than the actual ones). Of course feel free to disagree with me on this :)

Totally agree with you. I should have asked about the reason of using UUID in users and children models. I thought it is a requirement for every model ID field :P

events/factories.py
quyenlq: +1

events/models.py
```diff
@@ -0,0 +1,33 @@
...
+    time = models.DateTimeField(verbose_name=_("time"))
+    event = models.ForeignKey(
+        Event,
+        verbose_name=_("event"),
+        related_name="occurrences",
+        on_delete=models.CASCADE,
+    )
+
+    def __str__(self):
+        return f"{self.pk} {self.time}"
```
tuomas777: This class is missing translated `verbose_name` and `verbose_name_plural`.

venues/tests/test_models.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from ..factories import VenueFactory
+from ..models import Venue
+
+
+@pytest.mark.django_db
+def test_event_creation():
```
tuomas777: Copy-paste error in this name :)

events/models.py
tuomas777: Yeah, you are right. Not sure which one would be actually better, but I guess we should go with the diagram, so please add venue foreign key to occurrence instead :)

PR #28, opened 2020-01-14
KK-184 KK-183 | Add fetch APIs for event/occurrence/venue

quyenlq:


events/tests/test_api.py
```diff
@@ -0,0 +1,164 @@
...
+          }
+        }
+      }
+    }
+  }
+}
+"""
+
+OCCURRENCES_QUERY = """
+query Occcurences {
```
tuomas777: Typo.

quyenlq: > Good stuff! I don't think there is actually any need to restrict these queries to logged-in users only (sorry for practically non-existing specs for these queries), but I guess we can keep them like that for now, and change later if/when needed.
> 
> One thing, GraphQL queries don't normally include commas, so could you remove them from the test queries.

Yeah I think for now we can leave it like this, this will need to be updated anyway when we have more specs (filters, two types of admin ...)

PR #32, opened 2020-01-20
KK-198 | Add max depth for API queries

tuomas777:


kukkuu/urls.py
```diff
@@ -1,14 +1,23 @@
...
 from django.http import HttpResponse
 from django.urls import path
 from django.views.decorators.csrf import csrf_exempt
 from helusers.admin_site import admin
+from secure_graphene.depth import DepthAnalysisBackend
 
-from kukkuu import settings
 from kukkuu.views import SentryGraphQLView
 
+# the value for max depth is determined using the Stetson-Harrison method
```
charn: :cowboy_hat_face:

PR #37, opened 2020-01-23
KK-72 | Add event mutation

quyenlq:
Create event/venue/occurrences api
Override TranslatableModel to be able to update translation easier
Some unit tests
Improve admin interface a bit by adding inline form

### DIFF
--- a/common/models.py
+++ b/common/models.py
@@ -2,6 +2,8 @@ import uuid
 
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
+from parler.managers import TranslatableQuerySet as ParlerTranslatableQuerySet
+from parler.models import TranslatableModel as ParlerTranslatableModel
 
 
 class TimestampedModel(models.Model):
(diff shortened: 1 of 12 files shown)


events/schema.py
```diff
@@ -53,9 +57,80 @@ class Meta:
...
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Add validation
+        event = Event.objects.create_translatable_object(**kwargs)
+        # TODO: Add support for image upload
+        return AddEventMutation(event=event)
+
+
+class AddOccurrenceMutation(graphene.relay.ClientIDMutation):
+    class Input:
+        time = graphene.DateTime()
```
tuomas777: This should be required.

events/tests/test_api.py
```diff
@@ -212,3 +272,43 @@ def test_occurrence_query_normal_user(snapshot, user_api_client):
...
+        ADD_OCCURRENCE_MUTATION, variables=ADD_OCCURRENCE_VARIABLES
+    )
+    assert_permission_denied(executed)
+
+
+def test_add_occurrence_staff_user(snapshot, staff_api_client):
+    event = EventFactory()
+    venue = VenueFactory()
+    ADD_OCCURRENCE_VARIABLES["input"]["eventId"] = to_global_id("EventNode", event.id)
+    ADD_OCCURRENCE_VARIABLES["input"]["venueId"] = to_global_id("VenueNode", venue.id)
```
tuomas777: This is a bit bad, because it modifies `ADD_OCCURRENCE_VARIABLES` dict, so running this test will affect other tests as well. So you should (deep)copy the dict first. 

It could probably make sense to use pytest fixtures for variables instead of those dicts, but we can figure that out later :)

quyenlq: Thanks.  Didn't think of it because that variable has been used only in this test. I'll update it

common/models.py
```diff
@@ -19,3 +21,20 @@ class UUIDPrimaryKeyModel(models.Model):
...
+
+
+class TranslatableQuerySet(ParlerTranslatableQuerySet):
+    def create_translatable_object(self, **kwargs):
+        translations = kwargs.pop("translations")
+        obj = self.create(**kwargs)
+        for translation in translations:
+            language_code = translation.pop("language_code")
+            obj.create_translation(language_code=language_code, **translation)
+        return obj
```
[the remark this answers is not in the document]
quyenlq: Yeah. Although the outer functions calling this are all inside a transaction but I think it'd better to using transaction itself. Thanks for pointing it out :D

PR #40, opened 2020-01-28
Add support for upload image when creating event

quyenlq:


kukkuu/urls.py
```diff
@@ -24,3 +25,4 @@ def readiness(*args, **kwargs):
 
 
 urlpatterns += [path("healthz", healthz), path("readiness", readiness)]
+urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
tuomas777: This should be used only when `DEBUG` is `True`.

quyenlq: Aw shet, yeah,  fixed it

PR #46, opened 2020-02-03
Make get api public

quyenlq:
In order to support event page, backend needs to make the events/venues/occurrence API public, so user might be able to see events without login

### DIFF
--- a/events/tests/test_api.py
+++ b/events/tests/test_api.py
@@ -199,7 +199,6 @@ UPDATE_EVENT_MUTATION = """
 mutation UpdateEvent($input: UpdateEventMutationInput!) {
   updateEvent(input: $input) {
     event {
-      id,
       translations{
         name
         shortDescription
--- a/events/tests/test_api.py
+++ b/events/tests/test_api.py
@@ -304,10 +304,10 @@ mutation DeleteOccurrence($input: DeleteOccurrenceMutationInput!) {
 """
 
 
-def test_events_query_unauthenticated(api_client):
+def test_events_query_unauthenticated(api_client, event, snapshot):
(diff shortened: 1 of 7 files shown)


events/tests/test_api.py
```diff
@@ -304,10 +303,10 @@ def autouse_db(db):
 """
 
 
-def test_events_query_unauthenticated(api_client):
+def test_events_query_unauthenticated(api_client, event, snapshot):
```
charn: I guess the API now returns the same result when doing the query with a user or as unauthenticated user? Instead of having to somewhat similar tests we could pytest.parametrize to run the test with different api clients?

```

@pytest.mark.parametrize("client", [api_client, user_api_client])
def test_events_query(client, event, snapshot):
    ...
```

Not sure if the that actually works, but something you can consider if you want.

tuomas777: I don't think you can use fixtures in parametrize like that in plain pytest, you'll need an external package. Anyway, as these are probably needed in multiple tests, I think the most DRY solution for this would be to create a new parametrized fixture for the different api clients, something like:

```
@pytest.fixture(params=["api_client", "user_api_client", "guardian_api_client"])
def non_staff_api_clients(request):
    return request.getfixturevalue(request.param))


def test_x(non_staff_api_clients):
    non_staff_api_clients.execute(...)
```

quyenlq: Thanks for the comments. 
We discussed that those get API should be only visible to logged in users so I'm gonna close this PR
But yeah I'll refactor the fixtures soon when it's needed

quyenlq:
Close because we don't need this anymore

PR #50, opened 2020-02-07
KK-15 KK-256 | Add enrolment models and mutation

quyenlq:
- Add children-occurrences m2m relationship through enrolment


events/tests/test_api.py
```diff
@@ -547,3 +576,77 @@ def test_staff_publish_event(snapshot, staff_api_client, event):
...
+
+    assert "Child already joined this event" in str(executed["errors"])
+
+    random_child = ChildWithGuardianFactory()
+    enrolment_variables["input"]["childId"] = to_global_id("ChildNode", random_child.id)
+
+    executed = user_api_client.execute(
+        ENROL_OCCURRENCE_MUTATION, variables=enrolment_variables
+    )
+    assert "does not exist" in str(executed["errors"])
```
tuomas777: This is borderline bike-shedding, but it would be a bit nicer if this one big test that tests multiple things was divided into smaller tests, in this case I think there could be three different tests: one for a successful enrolment, one testing that you cannot enrol twice, and one testing that you cannot enrol someone else's child. Benefits: the tests would be easier to reason with and not dependent on each other.

Also as a future hint, there is `guardian_api_client` which might should be a bit easier to use than `user_api_client` for these tests that involve a guardian, of course no reason to change this stuff that is already implemented.

events/schema.py
```diff
@@ -132,6 +140,59 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        except Occurrence.DoesNotExist as e:
+            raise KukkuuGraphQLError(e)
+        try:
+            child = Child.objects.user_can_update(user).get(pk=child_id)
+        except Child.DoesNotExist as e:
+            raise KukkuuGraphQLError(e)
+        if child.occurrences.filter(event=occurrence.event).exists():
+            raise KukkuuGraphQLError("Child already joined this event")
+        occurrence.children.add(child)
+        enrolment = Enrolment.objects.get(child=child, occurrence=occurrence)
```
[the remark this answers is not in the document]
quyenlq: Updated, the initial idea was to use `.add` and `.remove` to trigger M2M `post_add` and `post_remove` signals. Unfortunately `django-admin` won't trigger those when creating enrolment through inline form so better stick with `Enrolment` `post_save` and `post_delete`signals.

events/tests/test_api.py
quyenlq: Okay I'll split it to smaller tests. I just thought it'd be nicer to test the whole flow without repeating duplicated data simulation

events/models.py
```diff
@@ -71,9 +72,45 @@ class Occurrence(TimestampedModel):
...
+        on_delete=models.CASCADE,
+        verbose_name=_("child"),
+    )
+    occurrence = models.ForeignKey(
+        Occurrence,
+        related_name="enrolments",
+        on_delete=models.CASCADE,
+        verbose_name=_("occurrence"),
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
```
tuomas777: This field should have a translatable verbose name.

events/notifications.py
```diff
@@ -2,19 +2,40 @@
...
 class NotificationType:
     EVENT_PUBLISHED = "event_published"
+    OCCURRENCE_ENROLMENT = "occurrence_enrolment"
+    OCCURRENCE_UNENROLMENT = "occurrence_unenrolment"
 
 
 notifications.register(NotificationType.EVENT_PUBLISHED, _("event published"))
+notifications.register(NotificationType.OCCURRENCE_ENROLMENT, _("occurrence enrolment"))
+notifications.register(
+    NotificationType.OCCURRENCE_UNENROLMENT, _("occurrence " "unenrolment")
```
tuomas777: A bit strange string formatting here :)

events/schema.py
```diff
@@ -132,6 +140,63 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         return DeleteEventMutation()
 
 
+class EnrolOccurrenceMutation(graphene.relay.ClientIDMutation):
+    class Input:
+        occurrence_id = graphene.GlobalID(
+            required=True, description="Occurrence id " "of event"
```
tuomas777: Ditto.

events/schema.py
```diff
@@ -132,6 +140,63 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            raise KukkuuGraphQLError("Child already joined this event")
+        enrolment = Enrolment.objects.create(child=child, occurrence=occurrence)
+
+        return EnrolOccurrenceMutation(enrolment=enrolment)
+
+
+class UnenrolOccurrenceMutation(graphene.relay.ClientIDMutation):
+    class Input:
+        occurrence_id = graphene.GlobalID(
+            required=True, description="Occurrence id " "of event"
```
tuomas777: Ditto.

quyenlq: Those must be conflicts between PyCharm auto format and black 💩

PR #55, opened 2020-02-17
KK-283 | Add enrolment validation

quyenlq:


events/schema.py
```diff
@@ -19,6 +23,15 @@
 EventTranslation = apps.get_model("events", "EventTranslation")
 
 
+def validate_enrolment(child, occurrence):
+    if child.occurrences.filter(event=occurrence.event).exists():
+        raise KukkuuGraphQLError("Child already joined this event")
+    if occurrence.enrolments.count() >= occurrence.event.capacity_per_occurrence:
+        raise KukkuuGraphQLError("Maximum enrolments created")
+    if occurrence.time < datetime.now(pytz.timezone(settings.TIME_ZONE)):
```
tuomas777: Django has helper [`timezone.now()`](https://docs.djangoproject.com/en/3.0/topics/i18n/timezones/#naive-and-aware-datetime-objects) for this.

quyenlq: Updated, thanks 👍. It'd be nicer if it generate timestamp with the TZ in `settings.py` instead of UTC tz but I guess it does really matter

tuomas777: Yes it would, for that you need `timezone.localtime(timezone.now())`. But yeah, timezones don't matter when comparing date times.

PR #59, opened 2020-02-19
KK-266 | Update README.md

quyenlq:


t0mim:
Looks good 👍

README.md
```diff
@@ -82,3 +95,15 @@ Or you can use [`pre-commit`](https://pre-commit.com/) to quickly format your co
...
 
 After that, formatting hooks will run against all changed files before committing
+
+## Jira
+
+- Link to Backlog:
+https://helsinkisolutionoffice.atlassian.net/secure/RapidBoard.jspa?rapidView=28&projectKey=KK&view=planning.nodetail&epics=visible&issueLimit=100
+
+- Active Sprint:
+https://helsinkisolutionoffice.atlassian.net/secure/RapidBoard.jspa?rapidView=28&projectKey=KK
```
tuomas777: Do we actually want to add links to non-public things here? I don't really like it :)

quyenlq: Hmn I thought these are public and can be accessed without permission but these aren't. Then I don't think it's necessary here, also the confluence page at the beginning should be removed as well

badrange:
@quyenlq Please change to using "Culture Kids" instead of Godchildren, we have a new and better English name for the project

PR #61, opened 2020-02-20
KK-285 | Update guardians filter when publish event

quyenlq:


events/tests/test_notifications.py
```diff
@@ -62,13 +62,12 @@ def notification_template_occurrence_unenrolment_fi():
...
-    guardians = GuardianFactory.create_batch(3, language="fi")
+    GuardianFactory(language="fi")
+    children = ChildWithGuardianFactory.create_batch(3)
     event_variables = deepcopy(PUBLISH_EVENT_VARIABLES)
     event_variables["input"]["id"] = to_global_id("EventNode", event.id)
     staff_api_client.execute(PUBLISH_EVENT_MUTATION, variables=event_variables)
-
-    assert len(mail.outbox) == len(guardians)
-    assert_mails_match_snapshot(snapshot)
+    assert len(mail.outbox) == len(children)
```
tuomas777: I suppose there is a reason why you cannot use a snapshot anymore?

quyenlq: I've experienced some random failed test while doing nothing. The assert snapshot here is not really stable, sometime the mails is not in the correct order, or the guardian factory generated different user emails (which is quite weird, I don't know why). --update-snapshot seems not work either, the next test after snapshot-update still failed

tuomas777: Ok. We can of course have just the amount assertion for now, I just think it is a bit too prone for false-positives. Ideally we would also check that we are sending the correct notifications to correct people.

Somewhat related to this, the combination of using faker data and snapshots in the tests turned out to be not ideal, the tests break pretty easily when something changes, and it is often not easy to say whether the functionality actually broke or was it just the test. So I think if/when we have time, we could check how we could improve the situation, I think replacing faker data with some more stable stuff like sequences could help. And if we do that, we could at the same time look into fixing this mail assert problem as well.

quyenlq: Totally agree! created [KK-306](https://helsinkisolutionoffice.atlassian.net/browse/KK-306) for that

PR #63, opened 2020-02-24
Add language support to venue and events

quyenlq:
Add normal translation fields to Venue and Event

### DIFF
--- a/venues/tests/test_api.py
+++ b/venues/tests/test_api.py
@@ -17,6 +17,9 @@ query Venues {
   venues {
     edges {
       node {
+        name
+        description
+        address
         translations {
           name
           description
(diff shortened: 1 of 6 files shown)


venues/tests/test_api.py
```diff
@@ -56,6 +59,9 @@ def autouse_db(db):
 VENUE_QUERY = """
 query Venue($id: ID!) {
   venue(id: $id){
+    name
+    description
+    address
```
tuomas777: Is there a reason why these are not using all the fields?

quyenlq: No specific reason at all, just too lazy to add because the snapshot might be too long

quyenlq:
> Looks good! As an addition, I think it would be good if the tests tested that we actually return different languages in the new fields :) If that requires much work, we can, of course, merge this first and add those tests later separately.

Updated a test case for that

PR #64, opened 2020-02-25
KK-309 | Update api to return event image absolute url

quyenlq:


events/tests/snapshots/snap_test_api.py
```diff
@@ -69,25 +69,14 @@
...
-                        "name": "Event test",
-                        "shortDescription": "Short desc",
-                    }
-                ],
-            }
+    "data": {"addEvent": {"event": None}},
+    "errors": [
+        {
+            "locations": [{"column": 7, "line": 12}],
+            "message": "Cannot return null for non-nullable field EventNode.image.",
```
tuomas777: Hmm, this doesn't seem right :)

quyenlq: Fixed, the field is not null in DB so `resolve_image` should return empty string instead of `None`

PR #66, opened 2020-02-26
Fix occurrences remaining capacity

quyenlq:
When querying occurrences from Event Node or Venue Node:

Screenshot 2020-02-26 at 14 56 06

Turns out `DjangoConnectionField` will discard the annotation declared in OccurrenceNode default queryset so we need to resolve it manually in its parent Node.

### DIFF
--- a/events/schema.py
+++ b/events/schema.py
@@ -65,6 +65,11 @@ class EventNode(DjangoObjectType):
             return info.context.build_absolute_uri(self.image.url)
         return ""
 
+    def resolve_occurrences(self, info, **kwargs):
+        return self.occurrences.annotate(
+            enrolments_count=Count("enrolments", distinct=True)
+        ).order_by("time")
+
 
 class EventConnection(Connection):
     class Meta:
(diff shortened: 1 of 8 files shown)


events/schema.py
```diff
@@ -80,11 +85,9 @@ class OccurrenceNode(DjangoObjectType):
     @classmethod
     @login_required
     def get_queryset(cls, queryset, info):
-        return (
-            queryset.annotate(enrolments_count=Count("enrolments"))
-            .select_related("event")
-            .order_by("time")
-        )
+        return queryset.annotate(
+            enrolments_count=Count("enrolments", distinct=True)
```
tuomas777: Why this `distinct` is needed?

quyenlq: Sometimes depends on the query, `enrolment_count` will be annotated more than one time(just graphene thing, I don't know what exact reason)  in both OccurrenceNode and the `resolve_occurrences` so distinct should be there to avoid duplicated count. It also avoids incorrect count when querying in a circle (event>occurrence>child>availableEvent>occurrence)

quyenlq: Sometimes depends on the query, `enrolment_count` will be annotated more than one time(just graphene thing, I don't know what exact reason)  in both OccurrenceNode and the `resolve_occurrences` so distinct should be there to avoid duplicated count. It also avoids incorrect count when querying in a circle (event>occurrence>child>availableEvent>occurrence)

tuomas777: Ok, thanks for the clarification :+1:

badrange: What about adding a code comment that explains why you added distinct here? 

You might wonder why you did it in a few months..

PR #69, opened 2020-02-27
Move MEDIA_ROOT to inside app folder

quyenlq:
Previously MEDIA_ROOT folder pointed to /var/media which wasn't accessible by docker `appuser`. The media folder should be inside /app/var/media.
Also an uwsgi `static-map` should be added to serve the images in production


kukkuu/settings.py
```diff
@@ -20,7 +20,7 @@
 env = environ.Env(
     DEBUG=(bool, False),
     SECRET_KEY=(str, ""),
-    MEDIA_ROOT=(environ.Path(), default_var_root("media")),
+    MEDIA_ROOT=(environ.Path(), environ.Path(checkout_dir("var"))("media")),
```
tuomas777: Hmm, not sure why this change is needed, isn't `environ.Path(checkout_dir("var"))` the same as `default_var_root` here?

quyenlq: For non-docker installation - yes, but for docker it will return /var/ instead of /app/var (https://github.com/City-of-Helsinki/kukkuu/pull/69/files/0d8ba0361eaf406a2300decb25952dd8288dbb46#diff-ddc70eb28cc2a5e5c42831bcfcc970d8R13-R18)
``` environ.Path(checkout_dir("var"))("media")``` will make sure it always return /app/var/media

PR #71, opened 2020-03-02
KK-315 | Fix duplicated availableEvent

quyenlq:


badrange:
This fixed the issue on my machine!

PR #74, opened 2020-03-04
Fix child events query to return only published events

quyenlq:
- Only staff user can query unpublished events (published_at=null) and occurrences of published event
    - Added queryset to filter events based on user
    - Added tests 
- Minor fix to `pastEvents` to return events which has no upcoming occurrences

### DIFF
--- a/children/schema.py
+++ b/children/schema.py
@@ -48,12 +48,15 @@ class ChildNode(DjangoObjectType):
 
     def resolve_past_events(self, info, **kwargs):
         # TODO: Only return events of the same project
-        return Event.objects.filter(occurrences__time__lt=timezone.now()).distinct()
+        return Event.objects.user_can_view(info.context.user).filter(
+            occurrences__time__lt=timezone.now()
+        )
 
     def resolve_available_events(self, info, **kwargs):
         # TODO: Only return events of the same project
         return (
-            Event.objects.filter(occurrences__time__gte=timezone.now())
+            Event.objects.user_can_view(info.context.user)
+            .filter(occurrences__time__gte=timezone.now())
             .distinct()
             .exclude(occurrences__in=self.occurrences.all())
         )
--- a/children/schema.py
+++ b/children/schema.py
@@ -48,8 +48,10 @@ class ChildNode(DjangoObjectType):
 
     def resolve_past_events(self, info, **kwargs):
         # TODO: Only return events of the same project
-        return Event.objects.user_can_view(info.context.user).filter(
-            occurrences__time__lt=timezone.now()
+        return (
+            Event.objects.user_can_view(info.context.user)
+            .exclude(occurrences__time__gte=timezone.now())
+            .distinct()
         )
 
     def resolve_available_events(self, info, **kwargs):
--- a/children/tests/test_api.py
+++ b/children/tests/test_api.py
@@ -600,13 +600,23 @@ def test_get_available_events(snapshot, guardian_api_client):
         relationship__guardian__user=guardian_api_client.user
     )
     variables = {"id": to_global_id("ChildNode", child.id)}
-    occurrences = OccurrenceFactory.create_batch(3, time=timezone.now())
+    # Unpublished occurrences
+    OccurrenceFactory.create(time=timezone.now())
+
+    # Published occurrences
+    occurrence = OccurrenceFactory.create(
+        time=timezone.now(), event__published_at=timezone.now()
+    )
+    OccurrenceFactory.create(time=timezone.now(), event__published_at=timezone.now())
+
+    # Past occurrences
     OccurrenceFactory.create_batch(
         3, time=datetime(1970, 1, 1, 0, 0, 0, tzinfo=pytz.timezone(settings.TIME_ZONE))
     )
-    EnrolmentFactory(child=child, occurrence=occurrences[0])
+
+    EnrolmentFactory(child=child, occurrence=occurrence)
     executed = guardian_api_client.execute(CHILD_EVENTS_QUERY, variables=variables)
-    assert len(executed["data"]["child"]["availableEvents"]["edges"]) == 2
+    assert len(executed["data"]["child"]["availableEvents"]["edges"]) == 1
     snapshot.assert_match(executed)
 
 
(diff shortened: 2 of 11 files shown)


children/tests/test_api.py
```diff
@@ -593,13 +593,25 @@ def test_get_available_events(snapshot, guardian_api_client):
         relationship__guardian__user=guardian_api_client.user
     )
     variables = {"id": to_global_id("ChildNode", child.id)}
-    occurrences = OccurrenceFactory.create_batch(3, time=timezone.now())
+    # Unpublished occurrences
+    OccurrenceFactory.create(time=timezone.now())
+
+    # Published occurrences
+    occurrence = OccurrenceFactory.create(
+        time=timezone.now(), event=EventFactory(published_at=timezone.now())
```
tuomas777: Just a tip in case you didn't know, I think you could do this also like `event__published_at=timezone.now()`

children/schema.py
```diff
@@ -48,12 +48,16 @@ def get_node(cls, info, id):
 
     def resolve_past_events(self, info, **kwargs):
         # TODO: Only return events of the same project
-        return Event.objects.filter(occurrences__time__lt=timezone.now()).distinct()
+        return Event.objects.filter(
+            occurrences__time__lt=timezone.now(), published_at__isnull=False
```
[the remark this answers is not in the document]
quyenlq: +1

quyenlq:
> Looks good, but we will actually need to limit the root `events` query to return only published events as well for UI users. But that can be done separately from this PR, as it requires some extra logic related to user permissions.
> 
> Also, this filtering for only published events is so common need, that IMO for DRY's sake it should be moved to an event [queryset method](https://docs.djangoproject.com/en/3.0/topics/db/managers/#creating-a-manager-with-queryset-methods), so we can do like `Event.objects.published()`everywhere.

+1 Nice catch! I'll refactor this abit

children/tests/test_api.py
```diff
@@ -625,19 +625,25 @@ def test_get_past_events(snapshot, guardian_api_client):
         relationship__guardian__user=guardian_api_client.user
     )
     variables = {"id": to_global_id("ChildNode", child.id)}
+
     # Unpublished occurrences
-    OccurrenceFactory.create(time=timezone.now())
+    OccurrenceFactory.create_batch(2, time=timezone.now(), event=EventFactory())
```
igordavydsson: does that create two occurences with two different unpublished events? do you want them to be occurences of the same event? or are they from the same event? 😅 not sure, how factory boy handles this.

igordavydsson: I am a bit confused by this test case 😅 , it generates quite a few occurences, but they are all connected to so many events. whenever you generate an occurence like this `OccurrenceFactory...`, without passing a pre-generated event, you get an occurence of a new event, if I understand it correctly.  so, only when you generate `past_occurrence_1` or do `OccurrenceFactory.create(time=timezone.now(), event=event)`, only then are these occurences connected to this event: `event = EventFactory(published_at=timezone.now())`. not sure if this is intentional, or desired logic, or am I just misunderstanding it. would the test pass if you just had this:
```python
def test_get_past_events(snapshot, guardian_api_client):
    child = ChildWithGuardianFactory(
        relationship__guardian__user=guardian_api_client.user
    )
    variables = {"id": to_global_id("ChildNode", child.id)}

    event = EventFactory(published_at=timezone.now())
    past_occurrence_1 = OccurrenceFactory.create(
        time=datetime(1970, 1, 1, 0, 0, 0, tzinfo=pytz.timezone(settings.TIME_ZONE)),
        event=event,
    )
    OccurrenceFactory.create(time=timezone.now(), event=event)

    EnrolmentFactory(child=child, occurrence=past_occurrence_1)

    executed = guardian_api_client.execute(CHILD_EVENTS_QUERY, variables=variables)
    # Still return enroled events if they are past events
    assert len(executed["data"]["child"]["pastEvents"]["edges"]) == 1
    snapshot.assert_match(executed)
```

igordavydsson: now that I am looking into your schema I am even more confused 😅 why do you guys have `available_events` and `past_events` on the `ChildNode`. why aren't those jusy their own queries inside `events.schema`? they aren't connected to the child node in any way, if I am reading it right.

igordavydsson: ah, okay, `available_events` is using `Child.occurences`, I see. but `past_events` just returns any event, that has occurences in the past. 🤔 is it desired behaviour?

igordavydsson: isn't it supposed to return events in the past that child was enroled to?

igordavydsson: I think you get `len(executed["data"]["child"]["pastEvents"]["edges"]) == 1` because only one of the several events you generate during this test case is published. others are not.

quyenlq: Well sorry for messing up with this test case :D The `pastEvents` is supposed to return event which has no upcoming occurrences (i.e all past occurrences) So in this test case there are:
- 2 recent occurrences of random unpublished events => filtered ❌
- 1 published event X has 3 occurrences, 1 from the past, 1 also from the past but enroled by a child, 1 from the current timestamp => filtered since still has one upcoming occurrence ❌
- 3 past occurrences of random unpublished events => filtered ❌
- 1 past occurrence of published event Y => not filtered  ✅

quyenlq: There are different types of occurrences due to the fact that for normal user unpublished events should not be shown in anycase

igordavydsson: no worries, this is partially because I am not that familiar with the business logic in this project. 🙂 

thanks for clarifying what's going on, the list you give helps a lot 👍 

I still find it a bit odd, that `ChildNode` has this field, which is not really related in any way to the `Child` model, but just returns `Events` that pass certain criteria. 😅 would make more sense IMO, if it was just a separate query within `events.schema.Query`, since the only parameter you need to filter them is user and it will be in the `info.context` anyway. is there any plan to filter those `pastEvents` based on the `Child` instance in the future? something similar to how you resolve `available_events` with `.exclude(occurrences__in=self.occurrences.all())`? in any case, changing this might be done in a separate PR.

quyenlq: You are right, at the moment there is no real relationship between children and events and we can just use events filter as separated queries. However in the future we will have a thing called `Project` which basically a group of children who born on the same year. Then that children can only join events of that project only. We don't want front-end to rewrite that part then so `availableEvents` and `pastEvents` a purpose

igordavydsson: ok, now it makes sense to me 🙂 👍

PR #75, opened 2020-03-05
Add make media directory to entrypoint

quyenlq:
Related to #69 
In `.prod/uwsgi.ini`, there is a static-map to map media folder from `/media` to `/app/var/media` but it wouldn't work if the `/app/var/media` doesn't exist before starting uwsgi. Added a `mkdir` in docker entrypoint to make sure the folder exists before mapping.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -3,6 +3,9 @@ set -e
 service cron start
 su - appuser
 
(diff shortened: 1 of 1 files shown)


quyenlq:
Closed since we are not going to use self-hosted image storages

PR #77, opened 2020-03-09
KK-339 | Add filter occurrences by venues

quyenlq:


events/filters.py
```diff
@@ -11,10 +12,13 @@ class OccurrenceFilter(django_filters.FilterSet):
     upcoming = django_filters.BooleanFilter(
         method="filter_by_upcoming", field_name="time"
     )
+    venue = django_filters.CharFilter(
```
igordavydsson: would `venue_id` be better name for this? to make it more descriptive 🙂 OTOH, `venue` could work too. your call.

events/filters.py
```diff
@@ -24,3 +28,7 @@ def filter_by_upcoming(self, qs, name, value):
         if value:
             return qs.filter(**{name + "__gte": timezone.now()})
         return qs
+
+    def filter_by_venue_global_id(self, qs, name, value):
+        venue_id = from_global_id(value)[1]
+        return qs.filter(venue=venue_id)
```
igordavydsson: would `filter(venue_id=venue_id) be more reliable? since you are not passing the `Venue` instance itself.

events/filters.py
quyenlq: That'd make sense +1. I'll update it and the qs params below

PR #80, opened 2020-03-09
Use uwsgi cron, remove crontab, revert Dockerfile and docker-entrypoint

quyenlq:
From #57: Using cronjob inside app container is not a good idea and requires many changes in Dockerfile & entrypoint.
Fortunately UWSGI has a cron-like interface that do exactly the same without any extra configuration so we should switch to using it


.prod/uwsgi.ini
```diff
@@ -9,3 +9,9 @@ gid = appuser
 master = 1
 processes = 2
 threads = 2
+cron = -1 -1 -1 -1 -1 /app/manage.py send_mail
+cron = 0 -1 -1 -1 -1 /app/manage.py retry_deferred
+cron = 20 -1 -1 -1 -1 /app/manage.py retry_deferred
+cron = 40 -1 -1 -1 -1 /app/manage.py retry_deferred
+cron = 0 0 -1 -1 -1 /app/manage.py purge_mail_log 7
+
```
igordavydsson: one line too many 😅

.prod/uwsgi.ini
```diff
@@ -9,3 +9,9 @@ gid = appuser
 master = 1
 processes = 2
 threads = 2
+cron = -1 -1 -1 -1 -1 /app/manage.py send_mail
```
igordavydsson: this is probably fine for now, when the service does not have that much traffic, but what if in the future sending all emails would take longer than 1 munite? 🤔 would every 5 minutes be a good enough frequency? or I am overthinking this? sending emails every minutes might be better from the UX POV for sure. 🙂

.prod/uwsgi.ini
```diff
@@ -9,3 +9,9 @@ gid = appuser
...
+cron = 0 -1 -1 -1 -1 /app/manage.py retry_deferred
```
igordavydsson: so, you want to run this every 20 minutes?
```
cron = -20 -1 -1 -1 -1 /app/manage.py retry_deferred
```

.gitlab-ci.yml
```diff
@@ -34,8 +34,10 @@ staging:
     K8S_SECRET_SECRET_KEY: "$GL_QA_DJANGO_SECRET_KEY"
     K8S_SECRET_SKIP_DATABASE_CHECK: 1
     K8S_SECRET_TOKEN_AUTH_AUTHSERVER_URL: "https://tunnistamo.test.kuva.hel.ninja/openid"
+    K8S_ILMOITIN_QUEUE_NOTIFICATIONS: 1
     K8S_SECRET_MAIL_MAILGUN_KEY: "$SECRET_MAILGUN_API_KEY"
     K8S_SECRET_MAIL_MAILGUN_DOMAIN: "mail.hel.ninja"
     K8S_SECRET_MAIL_MAILGUN_API: "https://api.eu.mailgun.net/v3"
     K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
     K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
+
```
igordavydsson: one line too many 👀

kukkuu/settings.py
```diff
@@ -42,7 +42,7 @@
     TOKEN_AUTH_ACCEPTED_SCOPE_PREFIX=(str, "kukkuu"),
     TOKEN_AUTH_REQUIRE_SCOPE_PREFIX=(bool, True),
     TOKEN_AUTH_AUTHSERVER_URL=(str, ""),
-    ILMOITIN_QUEUE_NOTIFICATIONS=(str, True),
+    ILMOITIN_QUEUE_NOTIFICATIONS=(str, False),
```
igordavydsson: so, for local dev env you want to disable queueing notifications?

quyenlq: I think that should be better. Since in local we probably don't use uwsgi but default django runserver to run the app so dev might don't want to setup cronjob in his local machine

.prod/uwsgi.ini
quyenlq: yeb, the idea is to retry deferred emails every 20 mins

.prod/uwsgi.ini
quyenlq: The thing is we also send notification for signup user and enrolment notification. I don't think 5 minutes delay is idea for such kinds of notification. I've tested with 1k emails and it tooks about 10-15 seconds so I won't think this should be a problem for now,  unless we have a huge email template

.gitlab-ci.yml
quyenlq: Fixed 😃

kukkuu/settings.py
igordavydsson: yes, sure! just checking that this was the goal here 🙂

.gitlab-ci.yml
```diff
@@ -34,8 +34,9 @@ staging:
     K8S_SECRET_SECRET_KEY: "$GL_QA_DJANGO_SECRET_KEY"
     K8S_SECRET_SKIP_DATABASE_CHECK: 1
     K8S_SECRET_TOKEN_AUTH_AUTHSERVER_URL: "https://tunnistamo.test.kuva.hel.ninja/openid"
+    K8S_ILMOITIN_QUEUE_NOTIFICATIONS: 1
     K8S_SECRET_MAIL_MAILGUN_KEY: "$SECRET_MAILGUN_API_KEY"
     K8S_SECRET_MAIL_MAILGUN_DOMAIN: "mail.hel.ninja"
     K8S_SECRET_MAIL_MAILGUN_API: "https://api.eu.mailgun.net/v3"
     K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
-    K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
+    K8S_SECRET_DEFAULT_FROM_EMAIL: "no-reply@hel.ninja"
```
igordavydsson: now newline at the end is missing 😬

.prod/uwsgi.ini
igordavydsson: ok 👍 let's go with the "every minute".

.prod/uwsgi.ini
```diff
@@ -9,3 +9,8 @@ gid = appuser
...
+cron = 20 -1 -1 -1 -1 /app/manage.py retry_deferred
+cron = 40 -1 -1 -1 -1 /app/manage.py retry_deferred
+cron = 0 0 -1 -1 -1 /app/manage.py purge_mail_log 7
```
igordavydsson: this is missing newline at the end too 😬 you had too empty lines before, now it's 0 empty lines, but it should be one

.prod/uwsgi.ini
igordavydsson: ok! you could use the line I suggested instead of the three you have 🙂

.prod/uwsgi.ini
quyenlq: Not sure what's going on with my IDE and github though -__-  I only see one empty line before so tried to delete one

.prod/uwsgi.ini
quyenlq: Ahhh. Yes! 😆Damn, why used three lines for this

PR #89, opened 2020-03-16
KK-366 | Add wc & facilities field to venue

quyenlq:


quyenlq:
https://github.com/City-of-Helsinki/kukkuu/pull/89/files#diff-430ce413128702abee2be78232777d67R18 Quick explains: This will improve the speed of test max enrolment since now there is only 50 enrolment created. Also, sometime Faker ran out of idea when trying to generate 1000 username (duplicated username generated)

quyenlq:
> Looks good! One minor whine: the commit message is somewhat misleading, as we are not actually renaming anything here.

Oh yah, not sure why I did that though

PR #102, opened 2020-03-31
KK-390 | Return occurrence & child from unenrolment mutation

quyenlq:


children/schema.py
```diff
@@ -31,8 +30,8 @@
 
 
 class ChildNode(DjangoObjectType):
-    available_events = relay.ConnectionField(EventConnection)
-    past_events = relay.ConnectionField(EventConnection)
+    available_events = relay.ConnectionField("events.schema.EventConnection")
```
quyenlq: This to avoid circular import in Event schema which import ChildNode from children schema

badrange:
I was able to solve the issue from KK-374 with the changes in this pr! 🎉

PR #104, opened 2020-04-02
KK-393 | Add null field validation when update obj

quyenlq:


common/utils.py
```diff
@@ -1,13 +1,16 @@
...
+from kukkuu.exceptions import DataValidationError
 from kukkuu.settings import REVISION
 
 
 def update_object(obj, data):
     if not data:
         return
     for k, v in data.items():
+        if v is None and not obj.__class__._meta.get_field(k).null:
+            raise DataValidationError("f{k} cannot be null.")
```
tuomas777, line 13: I guess this should be `f"{k} cannot be null."` :)

PR #112, opened 2020-04-23
KK-426 & KK-429 | Minor API enhancements mostly for Kukkuu admin UI

tuomas777:


events/schema.py
```diff
@@ -312,6 +313,7 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
             raise ObjectDoesNotExistError(e)
 
         occurrence = Occurrence.objects.create(**kwargs)
+        occurrence.enrolment_count = 0
```
quyenlq, line 316: What'd happened if we remove this line? Is it because the created object doesn't have annotated field?

tuomas777: Yes exactly, without that line `enrolmentCount` will get value `null`, which won't work because the field is not nullable, and we want the value to be `0` instead of `null` anyway. But good question, added a comment about that.

PR #113, opened 2020-05-05
KK-427 | Add name for projects

tuomas777:
Added translatable field "name" to Project model. It is populated with actual production values for year 2020 by a data migration.


quyenlq:
LGTM! Nice work 🥇

PR #118, opened 2020-05-28
KK-455 & KK-477 | Guardian email changes

tuomas777:
Made some changes to guardian email logic, please check the commit messages for details.

Also as a completely unrelated thing, fixed a link in CHANGELOG.


users/schema.py
```diff
@@ -8,16 +10,24 @@
...
 
 from .models import Guardian
 
 User = get_user_model()
 
 
+def validate_guardian_data(guardian_data):
+    if "email" in guardian_data:
+        try:
+            EmailValidator()(guardian_data["email"])
```
quyenlq, line 23: There is a `validate_email(<email>)` function from the same package which basically do the same thing but I think it looks nicer :D

users/tests/test_models.py
```diff
@@ -30,3 +30,21 @@ def test_guardian_with_children_creation(project):
...
+
+
+@pytest.mark.django_db
+def test_guardian_email_populating():
+    user = UserFactory(email="user@example.com")
+
+    guardian = Guardian.objects.create(user=user)
+    assert guardian.email == "user@example.com"
+
+    guardian.email = "guardian.example.com"
```
quyenlq, line 42: Is this a valid email address? I guess it shouldn't pass the validator

users/schema.py
tuomas777: That is indeed way nicer 👍

users/tests/test_models.py
tuomas777: Haha, good catch. Actually, the validation isn't run on model level, so if an email is set outside of admin UI and API, it is not validated. But I fixed the email to be a valid one so that the test isn't so confusing 🙂 

Whether the validation should be on model level is another question, perhaps, but I think that is out of the scope of this PR anyway, as other validations in Kukkuu work like this and should be all changed if we decide to do that.

quyenlq: Yeah, I knew that kukkuu validation doesn't run on model, but since it's an [EmailField](https://docs.djangoproject.com/en/3.0/ref/models/fields/#emailfield) so I guess there will be at least some validation there. Not sure why it didn't catch this :D

tuomas777: Yes one would assume that, but actually no validations are run by default, not even for EmailFields. As my previous test proved :D

quyenlq: Yah, just checked a bit, I was wrong. The model `.save()` indeed doesn't trigger any validation.

PR #121, opened 2020-06-04
KK-458 | Remove staff based permissions

tuomas777:
Changed viewing and admin CRUD mutations of Guardians (view only) Children (view only), Venues, Events and Occurrences to be allowed only for the corresponding project admins instead of all staff members like before.

Still to do later: Could use some new tests that check that a project admin of a different project cannot use the mutations. This is however not critical, because exploiting a possible hole would require a wrongdoer in the project admin group.

### DIFF
--- a/common/utils.py
+++ b/common/utils.py
@@ -1,8 +1,11 @@
 from django.db import transaction
+from graphene import Node
+from graphql_jwt.decorators import user_passes_test
+from graphql_jwt.exceptions import PermissionDenied
 from graphql_relay import to_global_id
 
 from kukkuu import __version__
-from kukkuu.exceptions import DataValidationError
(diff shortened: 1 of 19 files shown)


common/utils.py
```diff
@@ -30,3 +33,34 @@ def get_api_version():
...
+        yes_we_can = obj.can_user_administer(user)
+    except AttributeError:
+        raise TypeError(
+            f"{obj.__class__.__name__} model does not implement can_user_administer()."
+        )
+    if not yes_we_can:
+        raise PermissionDenied()
+
+
+def get_obj_from_global_id(info, global_id, expected_obj_type):
```
quyenlq, line 49: Nice utils function 👌  gonna copy it to Palvelutarjotin

PR #124, opened 2020-06-08
KK-456 | Add project filter to children, venues, events and occurrences queries

tuomas777:


events/filters.py
```diff
@@ -52,3 +54,7 @@ def filter_by_event_global_id(self, qs, name, value):
 
     def filter_by_occurrence_language(self, qs, name, value):
         return qs.filter(occurrence_language=value.lower())
+
+    def filter_by_project_global_id(self, qs, name, value):
```
quyenlq, line 58: It's really cool that for `venue` and `event` filters we don't need this function :D

PR #125, opened 2020-06-11
KK-497 | Enrolment attendance

tuomas777:
Added nullable boolean field `attended` to `Enrolment` model which can be
used to mark if the child actually attended the occurrence or not. The
value for the field can be set using new mutation
`SetEnrolmentAttendance(enrolmentId, attended)`.


events/schema.py
```diff
@@ -275,6 +279,42 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            enrolment = occurrence.enrolments.get(child=child)
+        except Enrolment.DoesNotExist:
+            raise ChildNotEnrolledError(
+                "The child has not been enrolled to the occurrence."
+            )
+
+        enrolment.attended = kwargs["attended"]
+        enrolment.save()
+
+        return UnenrolOccurrenceMutation(child=child, occurrence=occurrence)
```
quyenlq, line 315: I believe this should be `SetOccurrenceAttendance`? I also wonder if we should return the enrolment here since this is about enrolment mutation.

events/schema.py
```diff
@@ -275,6 +279,42 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+    @login_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        occurrence = get_obj_if_user_can_administer(
+            info, kwargs["occurrence_id"], Occurrence
+        )
+        child = get_obj_if_user_can_administer(info, kwargs["child_id"], Child)
+
+        if "attended" not in kwargs:
+            raise DataValidationError('"attendance" is required.')
```
quyenlq, line 303: Maybe `"attended" is required.` should make more sense here

events/schema.py
tuomas777: > I believe this should be `SetOccurrenceAttendance`?

Yes, it certainly should. I genuinely believe copy-paste functionality should be removed, nothing good ever comes out of it 😄 

 > I also wonder if we should return the enrolment here since this is about enrolment mutation.

Good question! That actually made me think about the whole mutation, and I decided to change it to be actually a mutation targeting enrolments, ie. instead of a child and an occurrence IDs you give it an enrolment ID. I think it is simpler and easier to use that way, 
let me know what you think 🙂

events/schema.py
tuomas777: Fixed.

events/schema.py
quyenlq: IMO that'd be better. I guess this mutation is mostly used in kukkuu-admin, where working with enrolment entities are more frequent (i.e enrolment ids are more likely to be  already queried) so this approach could be handy for FE rather than providing child-occurrence ids

PR #127, opened 2020-07-15
KK-509 | Occurrence cancelled notification

tuomas777:


events/models.py
```diff
@@ -132,6 +137,23 @@ class Meta:
     def __str__(self):
         return f"{self.pk} {self.time}"
 
+    def delete(self, *args, **kwargs):
+        if self.time >= now():
+            # this QS needs to be evaluated here, it would not work after the
+            # occurrence has been deleted
+            children = list(Child.objects.filter(enrolments__in=self.enrolments.all()))
```
quyenlq, line 144: Would just `self.children` work here?

tuomas777: Yes it would 🙂 . Changed.

PR #136, opened 2020-08-19
KK-526 | Add occurrence url to event notifications' contexts

tuomas777:


quyenlq (approved):
LGTM! 
Palvelutarjotin is gonna need the same thing soon. Thanks for the code, hehe

PR #141, opened 2020-09-11
KK-544 | Occurrence capacity override

tuomas777:
Events have `capacity_per_occurrence` which sets capacity for their all occurrences. Added nullable `capacity_override` to Occurrence which can be used to override the value coming from the event.

In the API side `OccurrenceNode` got two new fields, `capacity` and `capacityOverride`. Also changed occurrence `remainingCapacity` to show `0` instead of negative values in case there are more enrollments than capacity (which could happen in some special cases, like when capacity is reduced when there are already enrollments).


events/tests/test_notifications.py
```diff
@@ -110,9 +110,9 @@ def test_occurrence_enrolment_notifications_on_model_level(
     user_api_client,
     notification_template_occurrence_unenrolment_fi,
     notification_template_occurrence_enrolment_fi,
-    occurrence,
     project,
 ):
+    occurrence = OccurrenceFactory(id=74, time=now())
```
quyenlq, line 115: Is there any reason to specify the occurrence id here?

quyenlq (approved):
LGTM! Just a small question asked

events/tests/test_notifications.py
tuomas777: Yeah, otherwise the ID might change depending on which tests are run, for example like running `pytest` vs `pytest events`, making the test fail in one of them because it's snapshot is assuming the ID to be exactly some value.

A better fix would probably be to get the IDs out of the snapshots, which could be done at some point when there is too much time.

PR #144, opened 2020-09-20
KK-566 | Changes to event showing logic

tuomas777:
Implemented needed changes so that the UI can

* show enrolled events until half an hour after the occurrence has started
* show participated events as past events when the participated occurrence has passed

Implementation details in the commit messages.


charn (approved):
LGTM! :+1: :shipit:

PR #145, opened 2020-09-22
KK-400 & KK-402 | Free spot notification subscription model and initial API

tuomas777:
Added model `FreeSpotNotificationSubscription` model for handling subscriptions to free spot notifications, and API for viewing, adding and removing those subscriptions. Currently, there is no actual functionality related to those subscriptions.

The PR includes no tests, they will be implemented separately later because we need to ship this ASAP to not block frontend work.


events/schema.py
```diff
@@ -141,6 +145,18 @@ def resolve_enrolment_count(self, info, **kwargs):
...
+        child_id = get_node_id_from_global_id(kwargs["child_id"], "ChildNode")
+        try:
+            return (
+                Child.objects.user_can_view(info.context.user)
+                .get(pk=child_id)
+                .free_spot_notification_subscriptions.filter(occurrence_id=self.id)
+                .exists()
+            )
+        except Child.DoesNotExist as e:
+            raise ObjectDoesNotExistError(e)
```
quyenlq, line 158: Since this is a query without data modification, I'd prefer return `None` instead of throwing error here

tuomas777: Yeah good point!

Changed it to (hopefully) always return `None` when the ID doesn't match a child. It became a bit messy because of possible invalid IDs.

quyenlq (approved):
image

PR #146, opened 2020-09-25
KK-558 | Event reminder notification

tuomas777:
* normally a reminder will be sent 7 days prior to the occurrence

* if the enrolment is created on the same day when the reminder should
  be sent, or later than that, a reminder will not be sent

* when there is some kind of an abnormal situation and reminders haven't
  been sent 7 days prior to the occurrence, they can be also sent later
  closer to the event, at latest the day prior to the event


events/tests/test_notifications.py
```diff
@@ -192,3 +210,74 @@ def test_occurrence_cancelled_notification(
...
+
+
+@pytest.mark.django_db
+def test_occurrence_reminder_notification(
+    snapshot, notification_template_occurrence_reminder_fi, project,
+):
+    actual_now = now()
+
+    # time frozen so that the Enrolments will get created_at in the past
+    with freeze_time(actual_now - timedelta(days=8)):
```
charn, line 222: To make this test less likely to break if the setting is changed you could either read the setting (from the settings fixture) or use the settings fixture and just force `KUKKUU_REMINDER_DAYS_IN_ADVANCE` to an expected value.

```python
def test_occurrence_reminder_notification(... settings):
    settings.KUKKUU_REMINDER_DAYS_IN_ADVANCE = 7
```

charn: Looks like you have done it already. Please ignore this comment. :sweat_smile:

charn (approved):
LGTM! :+1: :shipit:

PR #150, opened 2020-09-30
Fix API version string

tuomas777:
Previously trying to get the API version was throwing an exception
when the revision could not be figured out, for example when running a
management script from outside the app's directory.


quyenlq (approved):
LGTM! Gotta steal it to Palvelutarjotin (aka beta Kultus) now :P

PR #155, opened 2020-10-14
KK-590 | Add language option "Other language"

tuomas777:
Added language option "Other language" which a guardian can select if a language she speaks at home is not available. Also updated the default languages.


quyenlq (approved):
LTGM! 👌

PR #156, opened 2020-10-15
Fix FreeSpotNotificationSubscription clean()

tuomas777:
FreeSpotNotificationSubscription.clean() was erroneously checking the remaining capacity of the occurrence on every clean, not just when creating a new subscription like it should have done.


quyenlq (approved):
Very nice!!!

PR #158, opened 2020-10-19
KK-579 & KK-599 | Manual messages model and API

tuomas777:
Implemented [model](https://helsinkisolutionoffice.atlassian.net/browse/KK-579) and [API](https://helsinkisolutionoffice.atlassian.net/browse/KK-599) for manual messages.

The PR does not include actual sending functionality nor tests so that the API can be shipped ASAP.


messaging/schema.py
```diff
@@ -0,0 +1,243 @@
...
+from kukkuu.exceptions import DataValidationError, MessageAlreadySentError
+
+from .models import Message
+
+logger = logging.getLogger(__name__)
+
+MessageTranslation = apps.get_model("messaging", "MessageTranslation")
+
+
+class OccurrenceList(graphene.ObjectType):
```
quyenlq, line 28: Is there any use of it below? :D

messaging/schema.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+
+        return AddMessageMutation(message=message)
+
+
+class UpdateMessageMutation(graphene.relay.ClientIDMutation):
+    class Input:
+        id = graphene.GlobalID()
+        translations = graphene.List(MessageTranslationsInput)
+        project_id = graphene.GlobalID(required=False)
```
quyenlq, line 128: Could be simply `ID()` then

messaging/schema.py
```diff
@@ -0,0 +1,243 @@
...
+                info, data.pop("event_id"), Event
+            )
+        else:
+            data["event"] = None
+        occurrences = [
+            get_obj_if_user_can_administer(info, occurrence_id, Occurrence)
+            for occurrence_id in data.pop("occurrence_ids", [])
+        ]
+
+        validate_event_and_occurrences(data["event"], occurrences)
```
quyenlq, line 111: Will this work as expected if `data["event"]` is `None`?

messaging/schema.py
```diff
@@ -0,0 +1,243 @@
...
+        log_text = f"user {info.context.user.uuid} deleted message {message}"
+        message.delete()
+
+        logger.info(log_text)
+
+        return DeleteMessageMutation()
+
+
+class Query:
+    message = relay.Node.Field(MessageNode)
```
charn, line 235: Are these queries admin only?

messaging/schema.py
tuomas777: Sloppy me :(

messaging/schema.py
tuomas777: That should not be possible because of the `if` starting from line 100.

messaging/schema.py
tuomas777: Yeah I guess it could, will change. I don't really get that graphene `GlobalID` vs `ID` anyway.

messaging/schema.py
tuomas777: Indeed they should be (but are not), will fix.

tuomas777:
Thanks for the comments, fixed.

messaging/schema.py
quyenlq: Well I should have asked clearer but does line 116 set data["event"] to `None`? I meant what if data["event"] is `None` during the validation

tuomas777: No, you shouldn't have, your question was great, but I was on crack. It should be fixed now.

quyenlq (approved):
looks great to me now 👌

PR #160, opened 2020-10-26
KK-582 & KK-604 | Manual messages functionality

tuomas777:
This PR adds functionality for sending and manual messages. API for that has been implemented already.

Manual messages' sending has quite a complex filtering rules specified [here](https://helsinkisolutionoffice.atlassian.net/browse/KK-582).

### DIFF
--- a/messaging/models.py
+++ b/messaging/models.py
@@ -1,14 +1,23 @@
-from random import random
+from functools import reduce
 
+from django.conf import settings
 from django.db import models
-from django.utils.timezone import now
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
+from django_ilmoitin.utils import send_mail
 from parler.models import TranslatedFields
+from parler.utils.context import switch_language
 from projects.models import Project
+from subscriptions.models import FreeSpotNotificationSubscription
 
+from children.models import Child
 from common.models import TimestampedModel, TranslatableModel, TranslatableQuerySet
-from events.models import Event, Occurrence
+from events.models import Enrolment, Event, Occurrence
 from users.models import Guardian
+
+
+class AlreadySentError(Exception):
+    pass
 
 
 class MessageQuerySet(TranslatableQuerySet):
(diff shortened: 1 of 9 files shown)


charn (approved):
LGTM, didn't spot anything critical. :+1: :shipit:

messaging/models.py
```diff
@@ -74,16 +83,94 @@ class Meta:
...
+
+        guardians = self.get_recipient_guardians()
+
+        self.sent_at = timezone.now()
+        self.recipient_count = len(guardians)
+        self.save(update_fields=("sent_at", "recipient_count"))
+
+        for guardian in guardians:
+            with switch_language(self, guardian.language):
+                if guardian.language in getattr(
```
quyenlq, line 98: I wonder would it be better to use the `send_notification`. How about creating a blank notification template with only `{{ body_text }}` inside?  I remember Venepaikka is using that one to handle custom message sending to end-user.

messaging/tests/test_sending.py
```diff
@@ -0,0 +1,198 @@
...
+    (
+        "no_event_or_occurrence",
+        "event",
+        "occurrence_tomorrow_1",
+        "occurrence_yesterday_1",
+    ),
+)
+@pytest.mark.django_db
+def test_message_sending_with_filters(snapshot, recipient_selection, event_selection):
+    """Ridiculously complex filtering test
```
quyenlq, line 71: It IS ridiculously complex 😂

messaging/models.py
```diff
@@ -74,16 +83,94 @@ class Meta:
...
-        self.sent_at = now()
-        self.save(update_fields=("recipient_count", "sent_at"))
+
+        if self.occurrences.exists():
+            occurrences = self.occurrences.all()
+            enrolments = enrolments.filter(occurrence__in=occurrences)
+            subscriptions = subscriptions.filter(occurrence__in=occurrences)
+        elif self.event_id:
+            occurrences = self.event.occurrences.all()
+            enrolments = enrolments.filter(occurrence__in=occurrences)
```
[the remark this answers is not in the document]
tuomas777: I don't really like how the two lines are repeated there, but because it is possible there are no occurrences, another option would need to be something like

        if self.occurrences.exists():
            occurrences = self.occurrences.all()
        elif self.event_id:
            occurrences = self.event.occurrences.all()
        else:
            occurrences = None
        if occurrences:
            enrolments = enrolments.filter(occurrence__in=occurrences)
            subscriptions = subscriptions.filter(occurrence__in=occurrences)

Which is even more lines 😄

messaging/models.py
tuomas777: Yeah good point. At first I was actually planning on doing that, but it requires the right kind of "dummy" template to exist in the db, which I don't like when it is not really needed. FWIW I think Ilmoitin should support this itself (sending a notification without a template).

messaging/models.py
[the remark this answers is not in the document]
quyenlq: ah ok, my bad, didn't know that there is a case `occurrences` is `None` 😥

quyenlq (approved):

tuomas777:
> LGTM, didn't spot anything critical. 👍 :shipit:

That is more than enough for us in Kukkuu! 🚀 💥

PR #166, opened 2020-11-09
KK-637 | Add recipient count to unsent messages

tuomas777:
Recipient count for an unsent message will be populated with the amount
of recipients there would be if the message was sent at that current
time.


quyenlq (approved):
Looks great! 👌
Would be great if there is a test case to test `get_recipient_count` returns `get_recipient_guardians().count()` in the messages query

PR #169, opened 2020-11-18
KK-647 | EventGroup GQL queries

tuomas777:
* EventGroup(Id) (get single EventGroup)
* EventsAndEventGroups(projectId) (get all EventGroups and all Events that don't belong to any group)

### DIFF
--- a/events/schema.py
+++ b/events/schema.py
@@ -25,7 +25,7 @@ from common.utils import (
     update_object_with_translations,
 )
 from events.filters import OccurrenceFilter
-from events.models import Enrolment, Event, Occurrence
+from events.models import Enrolment, Event, EventGroup, Occurrence
 from kukkuu.exceptions import (
     ChildAlreadyJoinedEventError,
     DataValidationError,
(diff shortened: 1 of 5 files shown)


events/schema.py
```diff
@@ -496,11 +561,33 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
     event = relay.Node.Field(EventNode)
+    event_group = relay.Node.Field(EventGroupNode)
     occurrence = relay.Node.Field(OccurrenceNode)
 
+    def resolve_events_and_event_groups(self, info, **kwargs):
+        event_qs = Event.objects.filter(event_group=None)
+        event_group_qs = EventGroup.objects.all()
+
+        if "project_id" in kwargs:
+            project_id = get_node_id_from_global_id(kwargs["project_id"], "ProjectNode")
```
charn, line 578: Maybe it's not really a big issue, but I think you can call this query explicitly with `project_id=null` which probably causes an exception from `get_node_id_from_global_id(None)`

charn (approved):
LGTM :+1: :shipit: One minor comment about the query argument, but not sure if that's even a real issue.

events/schema.py
tuomas777: Good point, fixed.

PR #172, opened 2020-11-24
Disable django-parler caching

tuomas777:
django-parler uses caching by default. That causes problems when there
are multiple Django processes and no proper cache backend in use
(which is exactly the situation in our QA and production envs ATM).


quyenlq (approved):
Nice catch! Good to know this can cause some problem for us 👌

PR #173, opened 2020-11-25
KK-627 | Add event group publishing functionality

tuomas777:
There are two ways an event group can be published:

1) using PublishEventGroup(id) mutation
2) using django admin UI publish event group action

Publishing an event group causes all of it's events to get published as
well. There is also a new notification that is sent when an event group
is published.


charn (approved):
LGTM! :+1: :shipit:

PR #174, opened 2020-12-01
KK-662 | Event group availability related API additions

tuomas777:
Details of the changes can be found in the commit messages.

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -93,6 +93,26 @@ class EventQueryset(TranslatableQuerySet):
     def unpublished(self):
         return self.filter(published_at__isnull=True)
 
+    def available(self, child):
+        """
+        A child's available events must match all of the following rules:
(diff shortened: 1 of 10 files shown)


events/models.py
```diff
@@ -93,6 +93,18 @@ def published(self):
     def unpublished(self):
         return self.filter(published_at__isnull=True)
 
+    def available(self, child):
+        child_enrolled_event_groups = EventGroup.objects.filter(
+            events__occurrences__in=child.occurrences.all()
+        )
+        return (
```
quyenlq, line 100: Would be great if there is a comment here 😄

quyenlq (approved):
Nicely done! 💯

events/models.py
tuomas777: I completely agree, added one 🙂

PR #177, opened 2020-12-04
KK-671 | Migrate tests to GitHub actions

tuomas777:
Added GitHub actions config for running tests. Pretty much copy-pasted from the glorious [berth-reservations](https://github.com/City-of-Helsinki/berth-reservations/blob/db4c4ffb926363625773d200917ba5e027fb1a97/.github/workflows/ci.yml).

Also fixed one issue in Kukkuu code which was making some of the tests randomly fail in some environments.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
+name: Continuous integration
+
+on:
+  push:
+    branches: [master]
```
a-rmz, line 5: We have this set to `master` since we're using `github-flow` and it's our default branch, but you might want to change it to either `[develop]` or `[develop, master]`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
...
+          path: ~/.pip-cache
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/requirements-dev.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        run: |
+          pip install -r requirements.txt -r requirements-dev.txt codecov && \
+          apt-get update && apt-get install -y build-essential
```
a-rmz, line 52: We had to add this to be able to install `gdal-bin` which was required for some geospatial stuff, so if you're not ussing that you could (1) get rid of this line and (2) use the "default" [python setup](https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python#starting-with-the-python-workflow-template) rather than the ubuntu-python docker image (same way it's done on the `coding-style` task).

tuomas777:
Thanks for the comments! Implemented the suggested changes.

charn (approved):
LGTM! :+1: :shipit: I'll copypaste this soon! :joy:

PR #178, opened 2020-12-13
KK-654 | Event "ready for publishing" field

tuomas777:
Ability to mark events as ready to be published as a part of an event group. This value needs to be true for all events of an event group when it is published.

Additions:

* `ready_for_event_group_publishing` boolean field to `Event` model

* `readyForEventGroupPublishing` to `AddEvent` and `UpdateEvent` mutations

* `PublishEventGroup` mutation returns error code `EVENT_GROUP_NOT_READY_FOR_PUBLISHING` when any of it’s events is not ready


events/admin.py
```diff
@@ -160,6 +163,12 @@ def save_model(self, request, obj, form, change):
         obj.events.set(form.cleaned_data["events"])
 
     def publish(self, request, queryset):
+        success_count = 0
         for obj in queryset:
-            obj.publish()
-        self.message_user(request, _("%s successfully published.") % queryset.count())
+            try:
+                obj.publish()
```
charn, line 169: So, if multiple event groups are published, but one of them raises an exception, I guess it still okay to publish the ones which do not cause an exception?

charn (approved):
AFAIK, LGTM! :+1: :shipit:

events/admin.py
charn: Since publishing will send notifications, this logic here is fine. 👍

tuomas777: Yup good point. Ideally if one fails the whole thing should fail, but event group publishing will send notifications, so that would not be really nice. Also in practice, it is highly unlikely multiple event groups will be ever published at the same time anyway, so this should be fine 🙂

PR #182, opened 2020-12-28
KK-688 | New permission system

tuomas777:
Started using Django's object permissions (using django-guardian) to handle Kukkuu admin user permissions. For now there are two `Project` permissions:

* `admin` which means the user can manage the project and do everything else BUT NOT publish events or event groups

* `publish` which means the user can publish events and event groups

In practice to use Kukkuu admin all admin users need the `admin` permission, ie. having just `publish` isn't enough.

There are a bunch of CRUD permissions Django creates by default, which just clutter the admin UI as we don't use them. Deleting all of those turned out to be quite a pain, so instead they are now just hidden from selection.


charn (approved):
AFAIK, LGTM! :+1: :shipit:

PR #183, opened 2020-12-30
KK-689 | Do not delete past enrolments when child is deleted

tuomas777:
Changed child field in Enrolment to be nullable. It will be set null for a past enrolment when its child is deleted. This change is made so that child deletion won't affect enrolment statistics in the past.


charn (approved):
:shipit:

PR #185, opened 2020-12-31
KK-690 | admin UI improvements

tuomas777:
Various smallish improvements to admin UI and a few other tweaks.

### DIFF
--- a/children/admin.py
+++ b/children/admin.py
@@ -13,6 +13,7 @@ class RelationshipInline(admin.TabularInline):
     extra = 0
     fields = ("guardian", "type", "created_at")
     readonly_fields = ("created_at",)
+    raw_id_fields = ("guardian",)
 
     def has_change_permission(self, request, obj=None):
         return False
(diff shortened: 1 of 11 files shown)


children/admin.py
```diff
@@ -45,21 +46,34 @@ class ChildAdmin(admin.ModelAdmin):
...
+    list_filter = ("project",)
+
+    def get_queryset(self, request):
+        return super().get_queryset(request).prefetch_related("guardians")
 
-    def get_project_year(self, obj):
-        return obj.project.year
+    def get_guardian(self, obj):
+        try:
+            return obj.guardians.all()[0]
```
charn, line 75: `obj.guardians.first()`? Or is there some reason why we need to have this try-except?

tuomas777: `.all()[0]` uses prefetched guardians. Could have added a comment about that.

charn (approved):
LGTM! :+1: :shipit:

PR #188, opened 2021-01-27
Define additional hostnames block

joonvena:


tuomas777 (changes requested):
Hmm, this doesn't look right. This seems to be from [events-helsinki-ui](https://github.com/City-of-Helsinki/events-helsinki-ui)(?).

PR #193, opened 2021-02-10
KK-715 | Fix User manager

tuomas777:
At least the `createsuperuser` management command was broken.


users/models.py
```diff
@@ -23,8 +23,13 @@ def possible_admins(self):
...
 
 
+# This is needed when using a custom User queryset
+class UserManager(OriginalUserManager.from_queryset(UserQuerySet)):
+    pass
+
+
 class User(AbstractUser):
-    objects = BaseUserManager.from_queryset(UserQuerySet)()
+    objects = UserManager()
```
charn, line 32: So I guess that separate class definition is required and `OriginalUserManager.from_queryset(UserQuerySet)` cannot be used here?

charn (approved):
LGTM! :+1: :shipit: I left a small comment, but as far as I know this looks awesome!

users/models.py
tuomas777: Yeah it seems so, without the class at least `makemigrations` fails with

```
ValueError: Could not find manager UserManagerFromUserQuerySet in django.db.models.manager.
Please note that you need to inherit from managers you dynamically generated with 'from_queryset()'.
```

charn: Okay then, all good here! :+1:

PR #198, opened 2021-02-12
KK-730 | Temporary auth fix needed with latest Tunnistamo

tuomas777:


charn (approved):
Awesome! :+1: :shipit:

PR #199, opened 2021-02-16
KK-726 | Add "single events allowed" flag to Project model

tuomas777:
When a Project has False in "single_events_allowed", it is not possible to create Events without an EventGroup using the API. Note that by design it is still possible to create such event using django admin UI, and if such event exists, it can be modified/deleted using the API. The default value for the flag is True so already existing Projects work like before.


events/schema.py
```diff
@@ -92,6 +93,9 @@ class EventNode(DjangoObjectType):
     class Meta:
         model = Event
         interfaces = (relay.Node,)
+        # fields = (
```
charn, line 96: :male_detective:

charn (approved):
Once that commented out code is removed, feel free to merge this! :+1: :shipit:

PR #200, opened 2021-02-17
KK-727 & KK-732 | Manage event groups perm

tuomas777:
Added new `Project` permission `manage_event_groups`, which is needed to be able to create, update and delete event groups using the GQL API.

While at it, also enabled the possibility to use model level `publish` permission ([KK-732](https://helsinkisolutionoffice.atlassian.net/browse/KK-732)).


charn (approved):
LGTM! :+1: :shipit:

PR #230, opened 2021-06-30
KK-765 | Allow empty ticket system URL for unpublished Ticketmaster events

tuomas777:
Changed occurrence ticket system URL to be required for Ticketmaster events only when the event is published. If some occurrence is missing the URL when the event is published, error code TICKET_SYSTEM_URL_REQUIRED will be returned from the API.


tuovinensanttu (approved):
AFAIK changes are looking good 👌

PR #231, opened 2021-08-04
KK-766 | Ticket system related mutations

tuomas777:
Added

*  ticket system type to event add and update mutation inputs. Ticket system cannot be changed if the event is already published.
* ticket system URL to occurrence add and update mutation inputs


charn (approved):
LGTM! :+1: :shipit:

PR #232, opened 2021-08-12
KK-769 | Change Event.capacity_per_occurrence nullable

tuomas777:
This change is done because external ticket system events don't need a capacity to be set. Capacity is still needed for internal ticket system events.


events/schema.py
```diff
@@ -414,7 +414,7 @@ class Input:
         translations = graphene.List(EventTranslationsInput)
         duration = graphene.Int()
         participants_per_invite = EventParticipantsPerInvite(required=True)
-        capacity_per_occurrence = graphene.Int(required=True)
+        capacity_per_occurrence = graphene.Int()
```
charn, line 417: I guess you could add a description here that it will be required if the ticket system is internal. Just for documentation purposes.

charn (approved):
LGTM! :+1: :shipit:

events/schema.py
tuomas777: Good idea, will do! 👍

PR #233, opened 2021-08-13
KK-767 | Ticket system password functionality

tuomas777:
Ticket system passwords are stored in their own model, and each password is for one child to one event.

At first when passwords are created, they aren't assigned to any children. When a password is needed for a child (`childPassword`field is requested via the GraphQL API), a new password is assigned for the child if the child does not have a password to the event yet. Later requests for a password for the child to the event will return the same password.


charn (approved):
LGTM! :+1:

PR #237, opened 2021-10-04
Kk-792 Fixes to the dev environment

nikomakela:
- Updated psycopg2 to psycopg2-binary. About the matter: https://stackoverflow.com/questions/33866695/error-installing-psycopg2-on-macos-10-9-5 and https://github.com/psycopg/psycopg2/issues/1293. Could not upgrade to latest, because it had issues with timezone. This one is also a patch upgrade to last one, so it should not need much testing either.
- Fix the CORS issues in Docker-compose env example configuration.
- Ignore vscode setup files.


requirements.txt
```diff
@@ -131,14 +131,14 @@ protobuf==3.11.3
     # via
     #   google-api-core
     #   googleapis-common-protos
-psycopg2==2.8.4
+psycopg2-binary==2.8.6
```
charn, line 134: Using the binary package removes the requirement to have postgres libraries installed for building the `psycopg2` package, which is convenient. However [psycopg2 documentation](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) states that especially production environments should only use the non-binary version.

nikomakela: Okay, nice catch! I'll try to address the problems in some other ways. Using psycopg2-binary was adviced by stack overflow.

nikomakela: Django 2.2 needs psycopg2 2.8. Mac needs the binary. The psycopg2-binary should be replacesd with psycopg2 when possible: https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary.

- https://github.com/biocommons/hgvs/issues/623
- https://github.com/psycopg/psycopg2/issues/674
- https://stackoverflow.com/questions/33866695/error-installing-psycopg2-on-macos-10-9-5
- https://github.com/psycopg/psycopg2/issues/1293

PR #238, opened 2021-10-11
KK-802 | Upgrade all dependencies

tuomas777:
Upgraded:

* Python to v3.9
* Django to v3.2
* Dev/CI Postgres to v11 (the version used in KuVa infra ATM)
* All packages to their latest versions


events/migrations/0023_alter_enrolment_attended.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("events", "0022_add_ticket_system_password"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="enrolment",
+            name="attended",
+            field=models.BooleanField(blank=True, null=True, verbose_name="attended"),
```
nikomakela, line 16: Shouldn't / couldn't this be default=False instead of blank=True and null=True?

events/models.py
```diff
@@ -467,7 +467,7 @@ class Enrolment(TimestampedModel):
         on_delete=models.CASCADE,
         verbose_name=_("occurrence"),
     )
-    attended = models.NullBooleanField(verbose_name=_("attended"))
+    attended = models.BooleanField(verbose_name=_("attended"), null=True, blank=True)
```
nikomakela, line 470: Shouldn't / couldn't this be default=False instead of blank=True and null=True?

nikomakela: I think this is something binary: you have attended or you have not. When answering null, the user or the system says that he does not want to answer or has not yet answered the question, but what is its benefit at this point?

PR #239, opened 2021-10-12
KK-803 | Fix free spot subscription handling in event group

tuomas777:
Enrolling to an occurrence should delete all free spot notification subscriptions from the child which are for the same event, or for an event that is in the same event group as the enrolled occurrence.

Before this fix, subscriptions were not deleted from the other events of the event group.

### DIFF
--- a/subscriptions/tests/test_models.py
+++ b/subscriptions/tests/test_models.py
@@ -3,9 +3,14 @@ from datetime import timedelta
 import pytest
 from django.utils.timezone import now
 from subscriptions.factories import FreeSpotNotificationSubscriptionFactory
-from subscriptions.tests.utils import assert_subscriptions
+from subscriptions.tests.utils import assert_child_has_subscriptions
 
-from events.factories import EnrolmentFactory, EventFactory, OccurrenceFactory
+from events.factories import (
+    EnrolmentFactory,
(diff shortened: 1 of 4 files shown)


subscriptions/tests/test_models.py
```diff
@@ -25,3 +30,38 @@ def test_enrolling_deletes_same_event_free_spot_subscriptions(
...
+    same_group_subscription = FreeSpotNotificationSubscriptionFactory(  # noqa: F841
+        child=child, occurrence=same_group_occurrence
+    )
+    other_group_subscription = FreeSpotNotificationSubscriptionFactory(
+        child=child, occurrence=other_group_occurrence
+    )
+
+    EnrolmentFactory(child=child, occurrence=occurrence)
+
+    assert_subscriptions(child, other_group_subscription)
```
nikomakela, line 67: Since I am still an outsider, I had to explore what this is doing. **Trivial** thing, but I would add comments on these lines so it would be clear what is being asserted. Or change the assert_subscriptions -function name, because it's not clear that assert_subscriptions -function compares child subscriptions to this other set of subscriptions.

Only a trivial thing, so can be approved like this also!

nikomakela (approved):
Looks good to me. Tested

PR #241, opened 2021-10-19
KK-731 | Improve authentication implementation

tuomas777:
This PR includes multiple authentication related improvements / fixes:

1.  Changed authentication implementation to use the new `RequestJWTAuthentication` class from django-helusers, which is meant to be used in projects that don't use Django REST Framework. This implementation is pretty much copy-pasted from https://github.com/City-of-Helsinki/open-city-profile.

    This change allowed us to get rid of dependencies for
      * drf-oidc-auth, and therefore the whole DRF
      * graphql_jwt (we still needed their `login_required` decorator, so
        that was copied to our codebase)

1. Fixed `PERMISSION_DENIED_ERROR`s to be not sent to Sentry
2. Implemented handling of helusers' `AuthenticationError`. It will result in `AUTHENTICATION_ERROR` error code being returned from our GraphQL API.

    If the exception is raised because of an expired JWT token, error code `AUTHENTICATION_EXPIRED_ERROR` will be returned instead. Currently helusers does not provide means to differentiate expired token from other authentication errors, so we needed some helusers internals related wizardry to make that happen.


nikomakela (approved):
I tested and debugged the authentication with basic auth (from /admin) and JWT (from /graphql) and the both seems to work ok. I didn't test the expiration, but saw the unit test for it and the implementation looks OK. Permission denied was thrown when session was closed in another browser tab.

I think the new JWT middleware could be set a bit higher in the hierarchy of middlewares, because it's now after message frameworks middleware, but that's a trivial thing:

```
MIDDLEWARE = [
    "django.middleware.security.SecurityMiddleware",
    "django.contrib.sessions.middleware.SessionMiddleware",
    "corsheaders.middleware.CorsMiddleware",
    "django.middleware.locale.LocaleMiddleware",
    "django.middleware.common.CommonMiddleware",
    "django.middleware.csrf.CsrfViewMiddleware",
    "django.contrib.auth.middleware.AuthenticationMiddleware",
    "django.contrib.messages.middleware.MessageMiddleware",
    "django.middleware.clickjacking.XFrameOptionsMiddleware",
    "kukkuu.middleware.JWTAuthentication",
]
```

Good job!

charn (approved):
LGTM! :+1: :shipit:

PR #246, opened 2021-12-17
KK-815 | Fix manual messages directed to "invited" to work with event groups

charn:
This PR makes the message recipient selection work correctly with event groups. A user is allowed to enroll to only one event in an event group. This means that if the user is enrolled to an event in an event group, the user doesn't have a valid invite to any other events in that group.

Also the message API is now checking that when the message is intended for "invited", the event and occurrence selections should remain empty.


tuomas777 (approved):
lgtm

PR #247, opened 2022-01-04
KK-828 | Feedback notification

tuomas777:
Added a new automatic notification which is meant to be sent after an occurrence has ended to ask for feedback from the participants. The notifications are sent by a management command which needs to be run periodically in staging and production. There is also a configurable additional delay with a default value of 15min.

The PR also contains some internal implementation changes to the already existing reminder notification, which make it more robust and consistent with the new feedback notification.


nikomakela (approved):
Good!

PR #254, opened 2022-01-24
Ignore authentication error like expired JWT from Sentry

nikomakela:
KK-833


nikomakela:
This might be something that we don't really want to do! I just thought that if the authenticationExpireError is ignored, why would the AuthenticationError still be raised in Sentry?

charn (approved):
LGTM! :+1: :shipit:

PR #256, opened 2022-02-03
Added pipeline yaml files

Vitals9367:


azure-pipelines-stageprod.yml
```diff
@@ -0,0 +1,46 @@
+# Continuous integration (CI) triggers cause a pipeline to run whenever you push 
+# an update to the specified branches or you push specified tags.
+trigger:
+  tags:
+    include:
+    - release-*tags/*
```
charn, line 6: Any reason why this is different in the UI related repos i.e. why is this not `- release-*`?

Vitals9367: Oh, my error. It has to be - release-*

Vitals9367: Should be fixed now.

nikomakela (approved):
looks good to me, but I have no experience with these

nikomakela:
Is there a good reason why the builds fails?

PR #257, opened 2022-02-07
Revert "Ignore authentication error like expired JWT from Sentry"

nikomakela:
Reverts City-of-Helsinki/kukkuu#254

Heard from @tuomas777 that these AuthenticationErrors are in monitoring in purpose, because some real guardians have repeatedly encountered this issue.


tuomas777 (approved):
👍

PR #259, opened 2022-02-25
Enrolment qrcode-ticket sent in mail

nikomakela:
When an enrolment is created, the guardian has get a mail in a text format. This pull request adds a QR-code to the mail's attachments. The QR-code is created from the ticket verification url and the enrolment's reference id, which is 5 chars code (lowercased alphabets) that is easy to write manually in the urls if needed.

The graphql-query:
```
  query VerifyTicket($referenceId: String!){
    verifyTicket(referenceId:$referenceId){
      occurrenceTime
      eventName
      validity
    }
  }
```

The Hashids is used to create the magic number that is used as a reference id to Enrolment instance. A salt is needed to prevent the malicious users to guess the magic numbers to tickets.

New mandatory settings
```
KUKKUU_HASHID_SALT=ULGd5YeRv6yVtvoj
KUKKUU_TICKET_VERIFICATION_URL=http://localhost:3000/ticket-verification-endpoint/{reference_id}
```


nikomakela:
I tried to test this locally and the mail was sent, but it never arrived. I guess this needs to be tested in the staging server (or another mail server is needed)

events/ticket_service.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+
+    enrolment_id = Enrolment.decode_reference_id(enrolment_reference_id)
+    try:
+        enrolment = Enrolment.objects.get(pk=enrolment_id)
+    except (Enrolment.DoesNotExist):
+        raise EnrolmentReferenceIdDoesNotExist(
+            "The decoded reference id does not match to any of the existing enrolments"
+        )
+    return (enrolment, enrolment.is_upcoming())
```
nikomakela, line 20: I'm not sure if this is enough to check. At least the UI should show the event name so wrong event ticket cannot be used multiple times!

events/schema.py
```diff
@@ -400,6 +401,9 @@ class Meta:
 
 
 class EnrolmentNode(DjangoObjectType):
+
+    reference_id = graphene.String(description="An unique encoded reference id")
+
```
nikomakela, line 406: Offering the enrolment reference id to authorized guardians, so the UI can recreate the qrcode.

requirements-dev.txt
```diff
@@ -50,6 +52,10 @@ mccabe==0.6.1
     # via flake8
 mypy-extensions==0.4.3
     # via black
+numpy==1.22.2
+    # via opencv-python
+opencv-python==4.5.5.62
+    # via -r requirements-dev.in
```
nikomakela, line 58: - [ ] Actually, I don't have any test with this yet. I should add one

nikomakela: The openCV is for PNGs only and I don't want to add svg to png converters, because they are not the main thing here. So, I removed the package that was needed after all.

events/models.py
```diff
@@ -520,6 +524,31 @@ class Meta:
...
+                ticket_qrcode_content,
+                "svg",
+            )
+            attachments.append(
+                (
+                    f"ticket-{self.reference_id}.svg",  # file name
+                    ticket_qrcode.decode(),  # content (decoded)
+                    "image/svg+xml",  # MIME-type
+                )
+            )
```
nikomakela, line 543: Haven't been able to fully test this yet, because the mails does not appear in my inbox. Not sure should the QR-code be a SVG or a PNG, decoded or not.

nikomakela: Moved this to a new notification service

events/tests/test_api.py
```diff
@@ -2290,3 +2290,114 @@ def test_event_ticket_system_password_not_own_child(guardian_api_client):
...
+        }
+      }
+      validity
+    }
+  }
+"""
+
+
+# NOTE: api_client is for anonymous users
+def test_verify_valid_ticket(api_client, snapshot):
```
charn, line 2313: I guess this query should not have access to `child` if it's unauthenticated

nikomakela: you are absolutely right! Could not figure out any other good solution than replace the whole enrolment object with 2 new fields: event name and occurrence time

kukkuu/urls.py
```diff
@@ -23,8 +23,8 @@
 
 urlpatterns = [
     path("admin/", admin.site.urls),
-    path(
-        "graphql",
+    re_path(
+        "graphql/?$",
```
charn, line 27: I think `r"^graphql/?$"` would be more the usual way.

nikomakela: "Fixed" :)

charn (approved):
LGTM! :+1: :shipit:

PR #261, opened 2022-03-01
Refactored notification service to support more formats and use PNG

nikomakela:
KK-831 Refactored the QR-code service and the notification service to better support multiple file formats for QR-codes and mail attachments. Changed the QR-code format of notification attachments from svg to png.


charn (approved):
LGTM! :+1: :shipit:

PR #265, opened 2022-03-04
SMS messaging

nikomakela:
KK-829

### Query

The API changes now includes a new field that tells which message protocol the message has been made for - the field name is “protocol“ and it’s in MessageNode. When creating a new message, the protocol-field must be included in the input.

### Filters

The messages query can now be filtered by a protocol, a list of occurrences (and with a project that has been there earlier too). 

### Database migration

A new char field / choice field called a protocol has been added. The default value for the protocol is EMAIL, which has been the main way to communicate and the only way before these changes.

Testing instructions and more here in the ticket https://helsinkisolutionoffice.atlassian.net/browse/KK-847


nikomakela:
Still a draft: This PR still needs the pipeline configurations. The API-token for the SMS notification service needs to be generated and configured to the pipeline environment variables.

nikomakela:
> Still a draft: This PR still needs the pipeline configurations. The API-token for the SMS notification service needs to be generated and configured to the pipeline environment variables.

Created the tokens and tested them locally: The test was a successful one.

kukkuu/settings.py
```diff
@@ -92,6 +96,13 @@
 ILMOITIN_TRANSLATED_FROM_EMAIL = env("ILMOITIN_TRANSLATED_FROM_EMAIL")
 ILMOITIN_QUEUE_NOTIFICATIONS = env("ILMOITIN_QUEUE_NOTIFICATIONS")
 
+if env.str("DEFAULT_SMS_SENDER"):
```
charn, line 99: Just a minor detail, but since `DEFAULT_SMS_SENDER` is already defined as string, this can be just `if env("DEFAULT_SMS_SENDER"):`

messaging/models.py
```diff
@@ -42,7 +35,15 @@ class Message(TimestampedModel, TranslatableModel):
             _("Subscribed to free spot notification"),
         ),
     )
-
+    EMAIL = "email"
+    SMS = "sms"
+    PROTOCOL_CHOICES = ((EMAIL, _("Email")), (SMS, _("SMS")))
+    protocol = models.CharField(
+        max_length=64,
```
charn, line 42: Just on observation that 64 seems like quite much when the current maximum length will be 5. Maybe 8 or 16? :sweat_smile:

.github/workflows/production.yml
```diff
@@ -61,6 +61,8 @@ jobs:
           K8S_SECRET_MAILER_EMAIL_BACKEND: "anymail.backends.mailgun.EmailBackend"
           K8S_SECRET_DEFAULT_FROM_EMAIL: "Kulttuurin kummilapset <noreply@hel.fi>"
           K8S_SECRET_ILMOITIN_TRANSLATED_FROM_EMAIL: "fi=Kulttuurin kummilapset <noreply@hel.fi>,en=Culture Kids <noreply@hel.fi>,sv=Kulturens fadderbarn <noreply@hel.fi>"
+          K8S_SECRET_NOTIFICATION_SERVICE_API_TOKEN: ${{ secrets.GH_NS_API_TOKEN_STABLE }}
```
charn, line 64: Should we add `TRANSLATED_SMS_SENDER` config here and staging?

charn (approved):
Few minor comments, but overall LGTM! :+1: :shipit:

.github/workflows/production.yml
nikomakela: Maybe if we get the translations. :) Checked from the Palvelutarjotin API and it's not configured there either.

Currently it works like this: 
```
    if language in getattr(settings, "TRANSLATED_SMS_SENDER", {}):
        sender = settings.TRANSLATED_SMS_SENDER[language]
    else:
        sender = settings.DEFAULT_SMS_SENDER
```
Maybe the else should be replaced with `if not sender`?

nikomakela: I will use the translations from the email
```
          K8S_SECRET_ILMOITIN_TRANSLATED_FROM_EMAIL: "fi=Kulttuurin kummilapset <noreply@hel.fi>,en=Culture Kids <noreply@hel.fi>,sv=Kulturens fadderbarn <noreply@hel.fi>"
```

PR #266, opened 2022-03-07
Added a feature for direct sending the message with addMessageMutation

nikomakela:
KK-829. The SMS messages should be able to be sent directly without any external add message step.


nikomakela:
This came up because of a note in this design image 
image

I personally think that the SMS and email should be handled the same way - It's the same message, just the communication protocol is different. If the SMS adding (/ sending) really should work differently than email, we have 2 options: 

1. allow the sending with addMessageMutation. This is done in this pull request
2. call sendMessageMutation from client right after addMessageMutation

I would actually prefer the latter one.

charn (approved):
Since `send_directly` avoids the extra step of having to make two consecutive API calls, I think this is fine. :+1: :shipit:

PR #267, opened 2022-03-07
Removed Translated SMS sender configuration from pipeline

nikomakela:
The SMS messages sending fails in the notification-service. Maybe the sender name is too long.

image


charn (approved):
LGTM! :+1: :shipit:

PR #268, opened 2022-03-09
KK-839 | Add enronlment limit and count to the API

charn:
Add project specific enrolment limit and child specific enrolment count to the API.
- ProjectNode.enrolmentLimit
- ChildNode.enrolmentCount


nikomakela (changes requested):
Code changes looks good, but a test fails! It also fails in my local! I also updated the snapshots and realised that a sms tests has been deleted. Wonder what is the reason for that 🤔

nikomakela (approved):

PR #271, opened 2022-03-14
KK-839 | Events occurrences have to be on the same year

charn:
This PR adds business logic which enforces all occurrences of an event to take place within the same year.

Also adds ChildNode.pastEnrolmentCount


events/schema.py
```diff
@@ -84,6 +85,26 @@ def validate_enrolment_deletion(enrolment):
         )
 
 
+def validate_occurrence(kwargs, occurrence: Occurrence = None):
```
Jokinen, line 88: Just pointing this out to make sure that it aligns with your naming conventions.

It seems like this function validates client input, i.e. there's no occurrence model instance yet. I would have thought that in a Django app model instances are the default case ("occurrence") and that client input is something different, for instance "occurrence input".

But if it is not a good convention for you to make that distinction, then please ignore.

Jokinen (approved):
Looks good to me 👍

PR #272, opened 2022-03-18
KK-839 | Add query for child's upcoming events and event groups

charn:
This PR adds an API to query all upcoming events and event groups for a child (`ChildNode.upcomingEventsAndEventGroups`). As opposed to the `availableEventsAndEventGroups` the new query also includes upcoming events to which the child might not be eligible to enroll.

`EventGroupNode` and `EventNode` include a new field `canChildEnroll(childId: ID)` which indicates if enrolling to the event is possible for the given child.

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -40,6 +40,15 @@ class EventGroupQueryset(TranslatableQuerySet):
     def published(self):
         return self.filter(published_at__isnull=False)
 
+    def upcoming(self):
+        return self.filter(events__occurrences__time__gte=timezone.now()).distinct()
+
+    def upcoming_with_leeway(self):
+        return self.filter(
+            events__occurrences__time__gte=timezone.now()
+            - timedelta(minutes=settings.KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY)
+        ).distinct()
+
 
 class EventGroup(TimestampedModel, TranslatableModel):
     translations = TranslatedFields(
@@ -80,6 +89,26 @@ class EventGroup(TimestampedModel, TranslatableModel):
 
     def can_user_publish(self, user):
         return user.can_publish_in_project(self.project)
+
+    def can_child_enroll(self, child) -> bool:
+        """Check if the child can enroll to an event in the event group."""
+
+        if occurrence := Occurrence.objects.filter(event__event_group=self).first():
+            # Need to have at least one occurrence
+            year = occurrence.time.year
+        else:
+            return False
+
+        if (
+            child.occurrences.filter(time__year=year).count()
+            >= child.project.enrolment_limit
+        ):
+            return False
+
+        if child.occurrences.filter(event__event_group=self).exists():
+            return False
+
+        return True
 
     def publish(self):
         unpublished_events = self.events.unpublished()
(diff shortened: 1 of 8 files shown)


events/schema.py
```diff
@@ -238,6 +240,14 @@ def resolve_ticket_system(self, info, **kwargs):
         # Event object is needed for resolving EventTicketSystem fields
         return self
 
+    def resolve_can_child_enroll(self: Event, info, **kwargs) -> Optional[bool]:
+        child_id = get_node_id_from_global_id(kwargs["child_id"], "ChildNode")
+        try:
+            child = Child.objects.user_can_view(info.context.user).get(id=child_id)
+        except Child.DoesNotExist:
+            return None
```
nikomakela, line 248: Shouldn't this raise an exception? At least the None does not add anything related to False, so I wonder why the None should ever be returned. A minor thing, but I would like to hear other opinions.

events/schema.py
```diff
@@ -301,6 +312,14 @@ def get_node(cls, info, id):
     def resolve_translations(self, info, **kwargs):
         return self.translations.order_by("language_code")
 
+    def resolve_can_child_enroll(self: EventGroup, info, **kwargs) -> Optional[bool]:
+        child_id = get_node_id_from_global_id(kwargs["child_id"], "ChildNode")
+        try:
+            child = Child.objects.user_can_view(info.context.user).get(id=child_id)
+        except Child.DoesNotExist:
+            return None
```
nikomakela, line 320: same thing as with `resolve_can_child_enroll`. Is the None there in purpose and thought through?

events/models.py
```diff
@@ -40,6 +40,15 @@ def user_can_view(self, user):
     def published(self):
         return self.filter(published_at__isnull=False)
 
+    def upcoming(self):
+        return self.filter(events__occurrences__time__gte=timezone.now()).distinct()
+
+    def upcoming_with_leeway(self):
+        return self.filter(
+            events__occurrences__time__gte=timezone.now()
+            - timedelta(minutes=settings.KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY)
```
nikomakela, line 49: What if the client could decide the amount of leeway minutes? No configuration needed then (or a default value could be added).

events/schema.py
```diff
@@ -977,7 +996,10 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
 class Query:
     events = DjangoFilterConnectionField(EventNode)
     events_and_event_groups = graphene.ConnectionField(
-        EventOrEventGroupConnection, project_id=graphene.ID()
+        EventOrEventGroupConnection,
+        project_id=graphene.ID(),
+        upcoming=graphene.Boolean(),
+        upcoming_with_leeway=graphene.Boolean(),
```
nikomakela, line 1002: Would it be better to let the client decide the amount of leeway? If so, then the Boolean should be changed to a number.

events/tests/snapshots/snap_test_api.py
```diff
@@ -524,6 +524,46 @@
...
+        }
+    }
+}
+
+snapshots["test_events_and_event_groups_query_upcoming_filter[True-False] 1"] = {
+    "data": {
+        "eventsAndEventGroups": {
+            "edges": [
+                {"node": {"__typename": "EventNode", "name": "Within leeway"}},
+                {"node": {"__typename": "EventNode", "name": "In the future"}},
```
nikomakela, line 550: Nice!

nikomakela (commented):
The code looks good. Wondering if the amount of leeway could be decided by the client and if the Child.DoesNotExists should raise an exception instead of returning None.

events/tests/snapshots/snap_test_api.py
charn: Yup, thought it's easier to check the snapshots for correct results.

events/models.py
charn: Using the strict leeway setting / boolean has/had the benefit that e.g. this result will be in sync with `childNode.pastEvents`.

Reason why I used boolean filters was simply that `OccurrenceNode` already had the same kind of `upcoming` and `upcoming_with_leeway` boolean filters and I wanted to keep this similar.

Another option (or additional filter) to using a strict or UI defined leeway value could be to just return events that are in-progress and upcoming. Currently `upcoming_with_leeway` returns future events and events which have started no longer than 30 minutes ago.

events/schema.py
charn: Raising an error sounds perfectly good. :+1:

nikomakela (approved):
good to go! 🚢

PR #273, opened 2022-03-22
KK-885 | Republish an event group

charn:
Event group can be republished if there are new unpublished events in the event group which are marked ready for event group publishing.


nikomakela (approved):
Good!

PR #274, opened 2022-03-31
 KK-892 KK-911 | Ticket QR code is valid for the whole day

charn:
This PR changes the QR code validation so that the ticket will appear to be valid for the whole day of the event occurrence.


nikomakela (approved):
Yes!

PR #275, opened 2022-04-04
KK-912 | Show event ticket for longer

charn:
The effect of this PR is that enrolled events are shown longer in the "you are enrolled" section of the UI, which should allow the user to access e.g. the event details and ticket for longer.

- Add new `upcomingWithOngoing` filter to occurrence queries (takes into account the event duration and leeway)
- `ChildNode.pastEvents` query takes into account the event duration (and leeway)
- Add a new `KUKKUU_DEFAULT_EVENT_DURATION` with default value of 120. This default value and end time calculation logic already existed for the feedback notification. In this PR we use the same logic for the `upcomingWithOngoing` filter and `ChildNode.pastEvents` query.
- Increase the default value of `KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY` from 30 to 60 minutes.
- Remove some unnecessary "upcoming with leeway" type of queries (not use by the UI).
- Change ticket validity checking logic. Ticket is now valid from 1 hour before the event, for the duration of the event (or the default duration) and for one hour after the event. This should be in line with how long the ticket is shown in the "you are enrolled" section.

**Note!** UI can/could/should now use `occurrences(upcomingWithOngoing: true)` instead of the `upcomingWithLeeway` filter.


events/models.py
```diff
@@ -369,11 +357,38 @@ def send_free_spot_notifications_if_needed(self):
...
+        qs = self.with_end_time()
+        return qs.filter(
+            end_time__gte=timezone.now()
+            - timedelta(minutes=settings.KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY)
+        )
 
     def in_past(self):
         return self.exclude(time__gt=timezone.now())
 
+    def with_end_time(self):
```
nikomakela, line 379: Good! Nice that Django ORM still casts this objects as Occurrence (and Enrolment).

nikomakela:
```
In [12]: e1 = Enrolment.objects.first()

In [13]: e2 = Enrolment.objects.with_end_time().first()

In [14]: e1.end_time
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-bfc454b61074> in <module>
----> 1 e1.end_time

AttributeError: 'Enrolment' object has no attribute 'end_time'

In [15]: e2.end_time
Out[15]: datetime.datetime(2001, 5, 31, 2, 49, 51, tzinfo=<UTC>)

In [16]: e1.__class__
Out[16]: events.models.Enrolment

In [17]: e2.__class__
Out[17]: events.models.Enrolment
```

nikomakela: Sure it might be a bad thing in some case when a new model instance is created from a dict of another.

nikomakela (approved):
Good!

PR #283, opened 2022-07-04
cronjob-is-broken-KK-754

terovirtanen:
Check if credentials file is exists


nikomakela (approved):
it sure looks clear at least :) Maybe I just let you try it 👍

nikomakela:
the PR name could be changed so it describes the change more

PR #289, opened 2022-09-01
KK-947 | Add AssignTicketSystemPasswordMutation & tests

karisal-anders:
## Description

AssignTicketSystemPasswordMutation(event_id, child_id) can be used to
assign a TicketSystemPassword to an event & child pair. If no free
passwords are avaiable then raises NoFreeTicketSystemPasswordsError,
if the TicketSystemPassword has already been assigned or another
TicketSystemPassword has already been assigned to the same child/event
pair then raises a TicketSystemPasswordAlreadyAssignedError.

Also:
 - Move event Exception class definitions to events/exceptions.py
 - Add TicketSystemPasswordAlreadyAssignedError KukkuuGraphQLError and
   map it to "TICKET_SYSTEM_PASSWORD_ALREADY_ASSIGNED_ERROR"
 - Change TicketSystemPassword.assign(self, event, child) function:
   - Remove event parameter as unnecessary
   - Raise PasswordAlreadyAssignedError if the assignment would result
     in breaking TicketSystemPassword's unique constraint for tuple
     (event, child)

## Context

[KK-947](https://helsinkisolutionoffice.atlassian.net/browse/KK-947)


karisal-anders:
Could @nikomakela or @tuomas777 please code review this pull request? Thank you! I welcome another pair of eyes/brains to discuss the [changes](https://github.com/City-of-Helsinki/kukkuu/pull/289/commits/9f69162c600fed4bc1fa01adb607cf6530ae1be3#diff-e22db1fd6b46c1799d0bb96df4d30720dba88c4ab310face6db5d9f9379a473c) made to the TicketSystemPassword.assign function's logic.

events/models.py
```diff
@@ -795,10 +795,23 @@ class Meta:
...
+            or (  # To keep unique constraint for tuple (event, child)
+                TicketSystemPassword.objects.exclude(pk=self.pk)
+                .filter(event=event, child=child)
+                .exists()
+            )
+        ):
             raise PasswordAlreadyAssignedError("The password is already assigned.")
 
         self.assigned_at = timezone.now()
+        self.event = event
```
tuomas777, line 814: I don't really understand why this function even has the parameter `event` 🙂 Unless I'm missing something, there shouldn't be any need to allow event changing here. A password should be tied to an event before it is assigned, and assigning a password should not change its event.

So I would just change the signature to `assign(self, child)`. Then the uniqueness checks should be also unnecessary.

tuomas777 (approved):
Great job 👍 🥇

PR #291, opened 2022-09-05
KK-950 | Implement GraphQL query for fetching a child's all internal and external enrolments

tuomas777:
Implemented query for fetching upcoming internal ticket system
enrolments (`Enrolment` model instances) and external ticket system
"enrolments" (assigned `TicketSystemPassword` instances). The enrolments
are sorted from earliest to latest where time is either the related
occurrence's time or the related event's latest occurrence's time
depending on which kind of enrolment it is.

The current purpose of this query is to provide a way for the UI to
fetch a child's all upcoming enrolments to "Tulevat tapahtumasi" (your
upcoming events) section in child profile view.

Example request:
```graphql
query ChildEnrolments {
  child(id: "Q2hpbGROb2RlOjJhYzY1MjE3LWNhZDktNDhlNi05ZWVhLTNjMWIwODhhOTIyZA==") {
    activeInternalAndTicketSystemEnrolments {
      edges {
        node {
          __typename
          ... on TicketmasterEnrolmentNode {
            createdAt
            event {
              name
            }
          }
          ... on EnrolmentNode {
            created_at
            occurrence {
              event {
		name
              }
            }
          }
        }
      }
    }
  }
}
```

Example response:
```json
{
  "data": {
    "child": {
      "activeInternalAndTicketSystemEnrolments": {
        "edges": [
          {
            "node": {
              "__typename": "EnrolmentNode",
              "createdAt": "2022-08-31T21:29:45.267373+00:00",
              "occurrence": {
                "event": {
                  "name": "Tapahtumien kuningas"
                }
              }
            }
          },
          {
            "node": {
              "__typename": "TicketmasterEnrolmentNode",
              "createdAt": "2022-08-11T12:09:13.819485+00:00",
              "event": {
                "name": "Toinen huikea tapahtuma"
              }
            }
          }
        ]
      }
    }
  }
}
```

### DIFF
--- a/children/tests/test_api.py
+++ b/children/tests/test_api.py
@@ -1499,3 +1499,130 @@ def test_test_upcoming_events_and_event_groups_yearly_enrolment_limit(
 
     for node in nodes:
         assert node["node"]["canChildEnroll"] is False
+
+
+CHILD_ACTIVE_INTERNAL_AND_TICKETMASTER_ENROLMENTS_QUERY = """
+query Child($id: ID!) {
+  child(id: $id) {
+    activeInternalAndTicketSystemEnrolments{
+      edges {
+        node {
+          ... on EnrolmentNode {
+            occurrence {
+              event {
+                name
+              }
+            }
+            __typename
+          }
+          ... on TicketmasterEnrolmentNode {
+            createdAt
+            event {
+              name
+            }
+            __typename
+          }
+        }
+      }
+    }
+  }
+}
+"""
+
+
+def test_active_internal_and_ticketmaster_enrolments(
+    snapshot,
+    guardian_api_client,
+    child_with_user_guardian,
+    future,
+    past,
+):
+    # unenrolled event, should not be returned
+    OccurrenceFactory(
+        event__name="INCORRECT",
+        event__published_at=now(),
+        time=future,
+    )
+
+    # unassigned Ticketmaster password, should not be returned
+    password_0 = TicketSystemPasswordFactory(
+        event=EventFactory(
+            published_at=now(),
+            ticket_system=Event.TICKETMASTER,
+            name="INCORRECT",
+        ),
+    )
+    OccurrenceFactory(event=password_0.event, time=future)
+
+    # enrolment in the past, should not be returned
+    enrolment = EnrolmentFactory(
+        child=child_with_user_guardian,
+        occurrence=OccurrenceFactory(
+            time=past,
+            event=EventFactory(published_at=now(), name="INCORRECT"),
+        ),
+    )
+    # add also one occurrence in the future, should not affect anything
+    OccurrenceFactory(event=enrolment.occurrence.event, time=future)
+
+    # enrolment in the future + 1 day, should be returned as the second enrolment
+    enrolment = EnrolmentFactory(
+        child=child_with_user_guardian,
+        occurrence=OccurrenceFactory(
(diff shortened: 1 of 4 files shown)


events/schema.py
```diff
@@ -471,6 +471,28 @@ def resolve_reference_id(self, info, **kwargs):
         return self.reference_id
 
 
+class TicketmasterEnrolmentNode(DjangoObjectType):
```
karisal-anders, line 474: Would user view restrictions like in `EnrolmentNode.get_queryset` function need to be implemented for TicketmasterEnrolmentNode?

events/schema.py
```diff
@@ -471,6 +471,28 @@ def resolve_reference_id(self, info, **kwargs):
...
+    class Meta:
+        model = TicketSystemPassword
+        interfaces = (relay.Node,)
+        fields = ("event", "created_at")
+
+    def resolve_created_at(self, info):
+        return self.assigned_at
+
+
+class InternalOrTicketSystemEnrolmentUnion(graphene.Union):
```
karisal-anders, line 486: There isn't much of a precedent in the current code base but thinking about the naming of graphene.Union type classes. `EventOrEventGroup` is the only other class which is a Union and it is a union of EventNode and EventGroupNode. Some naming conventions come to mind:

| A naming convention | Union₁ | Union₂ |
|---|---|---|
| Current  | InternalOrTicketSystemEnrolmentUnion | EventOrEventGroup |
| Use Union suffix for all | InternalOrTicketSystemEnrolmentUnion | EventOrEventGroupUnion |
| Use Node suffix for all | InternalOrTicketSystemEnrolmentNode | EventOrEventGroupNode |
| Don't use any shared suffix | InternalOrTicketSystemEnrolment | EventOrEventGroup |

A consistent naming convention would ease code comprehension IMHO.

What do you think?

events/schema.py
tuomas777: Good point.

In this particular case it shouldn't be strictly necessary yet, as those Ticketmaster enrolments can be accessed only through a child node, and that requires permissions. However, in my experience it is pretty easy to expose data in Graphene unintentionally, so I guess it is better to be safe than sorry and implement the restriction to the node's queryset already. That will be also consistent with the other nodes. The downside of course is that the resulting DB query will be a bit more complex, but I don't think that will be an issue in practice. 

So yeah, I will make that addition 🙂

events/schema.py
tuomas777: I think I will go with `Union` suffix all the way!

children/tests/test_api.py
```diff
@@ -1499,3 +1499,130 @@ def test_test_upcoming_events_and_event_groups_yearly_enrolment_limit(
...
+        child=child_with_user_guardian,
+        occurrence=OccurrenceFactory(
+            time=future + timedelta(days=1),
+            event=EventFactory(published_at=now(), name="2/4"),
+        ),
+    )
+    # add also one occurrence in the past, should not affect anything
+    OccurrenceFactory(event=enrolment.occurrence.event, time=past)
+
+    # enrolment in the future + 3 days, should be returned as the third enrolment
```
karisal-anders, line 1579: Comment out of sync i.e. third→fourth

karisal-anders (approved):
Some comment could be updated still and like you said combining some of the commits together etc but otherwise looks good! 👌

PR #293, opened 2022-09-16
Added the import ticket system passwords mutation

nikomakela:
KK-903. Added API test for ticket system password importing. Include errors field in import ticket system passwords mutation. Handle an empty list of passwords with an exception.


events/tests/test_api.py
```diff
@@ -2605,6 +2607,133 @@ def test_event_ticket_system_password_not_own_child(guardian_api_client):
...
+    snapshot.assert_match(executed)
+    assert (
+        executed["data"]["importTicketSystemPasswords"]["event"]["name"] == event.name
+    )
+    # Passwords should contain only the added passwords
+    assert executed["data"]["importTicketSystemPasswords"]["passwords"] == new_passwords
+    errors_as_text = str(
+        executed["data"]["importTicketSystemPasswords"]["errors"]["passwords"]
+    )
+    assert all([p in errors_as_text for p in existing_passwords])
```
karisal-anders, line 2676: You could also add the other way around:
```python
assert all([p not in errors_as_text for p in new_passwords])
```

events/tests/test_api.py
```diff
@@ -2605,6 +2607,133 @@ def test_event_ticket_system_password_not_own_child(guardian_api_client):
...
+        variables={"input": {"eventId": get_global_id(event), "passwords": None}},
+    )
+    assert_match_error_code(executed, GENERAL_ERROR)
+
+    # No passwords
+    executed = project_user_api_client.execute(
+        IMPORT_TICKET_SYSTEM_PASSWORDS_MUTATION,
+        variables={"input": {"eventId": None, "passwords": ["secret-coupon"]}},
+    )
+    assert_match_error_code(executed, GENERAL_ERROR)
```
karisal-anders, line 2699: Comment and code are swapped.

events/tests/test_api.py
```diff
@@ -2605,6 +2607,133 @@ def test_event_ticket_system_password_not_own_child(guardian_api_client):
...
+def test_import_ticket_system_passwords_event_not_exist(project_user_api_client):
+    """The event where the passwords are imported does not exist"""
+    event = EventFactory(ticket_system=Event.TICKETMASTER)
+    event_id = get_global_id(event)
+    event.delete()
+    executed = project_user_api_client.execute(
+        IMPORT_TICKET_SYSTEM_PASSWORDS_MUTATION,
+        variables={"input": {"eventId": event_id, "passwords": None}},
+    )
+    assert_match_error_code(executed, GENERAL_ERROR)
```
karisal-anders, line 2719: Maybe provide passwords so the problem is not related to omitting them.

kukkuu/exceptions.py
```diff
@@ -124,5 +124,9 @@ class TicketSystemPasswordAlreadyAssignedError(KukkuuGraphQLError):
     """
 
 
+class TicketSystemPasswordNothingToImportError(KukkuuGraphQLError):
+    """Empty list of passwords given"""
```
tuomas777, line 128: Not a huge deal and does not really matter in practice that much, but in principle, I think these specific errors should be reserved mostly for situations where the client needs to be able to identify the error programmatically (yeah, not all Kukkuu's current errors follow that either 😄)

Here sending an empty list is clearly an error in the client's code, so the returned error is just for developers, and the errorcode in it does not matter at all. There is a general `DataValidationError` which is used in this kind of situations elsewhere in the code, so I would rather use that with some kind of an error message describing the error.

events/schema.py
```diff
@@ -524,6 +527,97 @@ class EventTicketSystemInput(graphene.InputObjectType):
...
 
 
+class ImportTicketSystemPasswordErrorType(graphene.Scalar):
+    @staticmethod
+    def serialize(errors):
+        if isinstance(errors, dict):
+            if errors.get("__all__", False):
+                errors["non_field_errors"] = errors.pop("__all__")
+            return camelize(errors)
+        raise Exception("`errors` should be dict!")
```
tuomas777, line 536: Maybe `TypeError` could be a bit more appropriate?

nikomakela:
an alternative solution with create_bulk #294

events/schema.py
nikomakela: yep, refactored this

events/schema.py
```diff
@@ -524,6 +527,97 @@ class EventTicketSystemInput(graphene.InputObjectType):
...
+        )
+
+    @classmethod
+    def _create_batch(cls, passwords_batch: List[TicketSystemPassword]):
+        imported_passwords: list[TicketSystemPassword] = []
+        passwords_with_errors: list[str] = []
+        for obj in passwords_batch:
+            try:
+                # Use atomic to create a new transaction
+                with transaction.atomic():
```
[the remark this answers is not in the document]
nikomakela: When the loop is processed, if the previous object had an integrity issue for example, the transaction will be locked for the next object. I tried to find a solution to end the transaction when an exception occurs, but couldn't find any. Opening a new transaction was advised. We do that also in other parts of the code.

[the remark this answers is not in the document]
tuomas777: Right, that seems remotely familiar now, thanks for the explanation. I have to say a shorter version of that would be a way more helpful comment there instead of the one just stating the obvious 🙂

common/schema.py
```diff
@@ -16,3 +18,27 @@ def set_obj_languages_spoken_at_home(info, obj, language_global_ids):
         obj.languages_spoken_at_home.add(
             get_obj_from_global_id(info, language_global_id, Language)
         )
+
+
+class ErrorType(graphene.ObjectType):
+    """
+    A geric error type that can be used to add errors inside the data,
```
karisal-anders, line 25: Typo: geric → generic

common/schema.py
```diff
@@ -16,3 +18,27 @@ def set_obj_languages_spoken_at_home(info, obj, language_global_ids):
...
+    message = graphene.String(required=True)
+    value = graphene.String(required=True)
+
+    # @staticmethod
+    # def serialize(errors):
+    #     if isinstance(errors, dict):
+    #         if errors.get("__all__", False):
+    #             errors["non_field_errors"] = errors.pop("__all__")
+    #         return camelize(errors)
+    #     raise Exception("`errors` should be dict!")
```
karisal-anders, line 44: Leftover comment? Just remove?

events/tests/snapshots/snap_test_api.py
```diff
@@ -754,13 +755,23 @@
...
-                    ["Could not import password", "asd"],
-                    ["Could not import password", "xyz321"],
-                ]
-            },
+            "errors": [
+                {
+                    "field": "passwords",
+                    "message": "Could not import password",
+                    "value": "123",
+                },
```
karisal-anders, line 763: 👍 At least for me this reads clearer than the previous version.

events/tests/test_api.py
```diff
@@ -2670,10 +2674,12 @@ def test_import_ticket_system_passwords_errors_with_integrity_errors(
     )
     # Passwords should contain only the added passwords
     assert executed["data"]["importTicketSystemPasswords"]["passwords"] == new_passwords
-    errors_as_text = str(
-        executed["data"]["importTicketSystemPasswords"]["errors"]["passwords"]
+    assert all(
+        [
+            p in str(executed["data"]["importTicketSystemPasswords"]["errors"])
+            for p in existing_passwords
+        ]
```
karisal-anders: I'd suggest not using `p in str(...)` here as there are more type safe and correct ways, e.g. something like (Didn't actually try if this works):
```python
error_values = [error["value"] for error in executed["data"]["importTicketSystemPasswords"]["errors"]]
assert all(p in error_values for p in existing_passwords)
assert all(p not in error_values for p in new_passwords)
```

karisal-anders (commented):
Left some comments, please take a look at them, thank you!

events/tests/test_api.py
nikomakela: it will have a different format when I get the damn CI to handle the snapshots the same way as my local env does.

kukkuu/exceptions.py
nikomakela: Usually a problem is that there are too generic errors or no proper logging at all. Because there are pretty many specific error codes 1 does not hurt much more :) I too think that there are too specific codes, but in passwords handling there was followed this pattern already so...

reports/api.py
```diff
@@ -125,11 +125,15 @@ def get_languages_spoken_at_home(self, obj: Child) -> List[Optional[str]]:
...
+    queryset = (
+        Child.objects.exclude(guardians=None)
+        .prefetch_related(
+            Prefetch(
+                "guardians__languages_spoken_at_home",
+                queryset=Language.objects.order_by("alpha_3_code"),
+                to_attr="prefetched_languages_spoken_at_home",
+            ),
+        )
+        .order_by("last_name", "first_name")
```
karisal-anders, line 137: Is this a fix for [KK-953](https://helsinkisolutionoffice.atlassian.net/browse/KK-953)?

nikomakela: most probably

nikomakela: its here also https://github.com/City-of-Helsinki/kukkuu/pull/293/files#diff-58d75c954f04eb71a49cfbd1338dca4b196bc590b6e2ae2b424f1f23416519bfR24

karisal-anders: Yep, looks like the data in the bug report KK-953.

karisal-anders (approved):
Looks 👌 to merge!

PR #294, opened 2022-09-16
Use bulk_create to import ticket system passwords

nikomakela:
KK-903. Alternative implementation for ticket system passwords importing. Using the create_bulk -command, the insert-process will be a single request, but the errors other than integrity issues will just raise an error that we cannot handle.

NOTE: This does not improve much and the for-loop-importing should not be an issue in Kukkuu at all. Just wanted to try this and ask for your thoughts

### DIFF
--- a/events/schema.py
+++ b/events/schema.py
@@ -5,7 +5,7 @@ from typing import List, Optional
 import graphene
 from django.apps import apps
 from django.core.exceptions import PermissionDenied, ValidationError
-from django.db import IntegrityError, transaction
+from django.db import transaction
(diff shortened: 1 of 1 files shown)


karisal-anders (approved):
Looks good enough for me! The good enough has to do with the Kukkuu's probable use cases, probability of simultaneous deleting/inserting of passwords for the same event in Kukkuu environment etc, I don't think there's a need to needlessly engineer this further to try to account for all kinds of different simultaneous use cases if they are not probable.

PR #300, opened 2022-10-03
KK-961 | External ticket system event URL

tuomas777:
Added ticket system URL field to Event model and API. The idea is that it will replace the current ticket system URL field in occurrences, but the latter is also kept there for now to not break the UIs.

Also removed the ability to change ticket system type on event update, because that ability wasn't needed or used, and it would have made the event update logic related to the new field more complex.


nikomakela (approved):
Looking good! Approved

PR #302, opened 2022-10-13
KK-945 && KK-971 | Implement end time handling for external ticket system events

tuomas777:
External ticket system events don't have occurrences, so we cannot use those to determine the event's end time as is done for normal events. Event end time is needed with queries `child.activeInternalAndTicketSystemEnrolments`, `child.upcomingEventsAndEventGroups` and `child.pastEvents`, basically "Your upcoming events", "Event invitations" and "Past events" sections in the child profile view in the UI (pictured).

To tackle the situation, new field `ticket_system_end_time` is added to `Event` model. That field should be populated for external ticket system events. Changes required to the API queries' implementations are also included in this PR.

Screenshot 2022-10-17 at 9 44 32


children/schema.py
```diff
@@ -128,18 +128,54 @@ def resolve_past_enrolment_count(self: Child, info, **kwargs):
...
+          * internal ticket system Events the user has enrolled and the end time of the
+            Enrolment's Occurrence is in the past
+          * external ticket system Events the user has a password to and the Event's
+            ticket_system_end_time is in the past
+
+        To be considered as being in the past an end time needs to be actually
+        KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY mins in the past.
         """
-        events = self.project.events.user_can_view(info.context.user).published()
+        from events.schema import OccurrenceNode, TicketmasterEnrolmentNode  # noqa
```
karisal-anders, line 140: Are you importing here to avoid circular imports or is there some other reason?

children/schema.py
```diff
@@ -128,18 +128,54 @@ def resolve_past_enrolment_count(self: Child, info, **kwargs):
...
-            - timedelta(minutes=settings.KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY)
+            end_time__lt=past_enough,
+            event__in=published_events,
+        ).select_related("event")
+
+        past_ticket_system_passwords = (
+            self.ticket_system_passwords.filter(
+                (
+                    ~Q(event__ticket_system=Event.INTERNAL)
+                    & Q(event__ticket_system_end_time__lt=now)
```
karisal-anders, line 160: now → past_enough?

children/schema.py
```diff
@@ -264,7 +300,7 @@ def validate_child_data(child_data):
     if "birthdate" in child_data:
         birth_year = child_data["birthdate"].year
         if (
-            child_data["birthdate"] > localtime(now()).date()
+            child_data["birthdate"] > localtime(timezone.now()).date()
```
karisal-anders, line 316: django.utils seems to have `localdate` (See [docs](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.timezone.localdate)) which should give today intepreted in local time if nothing is passed to it.

children/schema.py
tuomas777: That is a good question, and I actually started implementing it like that first. However, there are multiple reasons why I decided to go with now instead. Explaining the whole thing would not be easy without a wall of text, so lets rather have an actual discussion about that elsewhere if needed. However, doing that either way should not really matter at all in practice, so it is questionable how much time we want to spend here 😄

children/schema.py
tuomas777: True, changing to that 👍

children/tests/test_api.py
```diff
@@ -1063,6 +1078,134 @@ def test_get_past_events(
     )
 
 
+def test_get_past_events_including_external_ticket_system_events(
+    snapshot, guardian_api_client, child_with_user_guardian, project, venue
+):
+    variables = {"id": to_global_id("ChildNode", child_with_user_guardian.id)}
+    duration = 30
```
karisal-anders, line 1085: `duration_mins` or `duration_minutes` would IMHO be clearer than `duration`. Event.duration is already named badly so it starts there. So, maybe that could be renamed there too although that's not necessary in this PR.

tuomas777: Yeah agree. And whether it actually should be minutes at all is another good question 😄 However, changing just the field's name would require also changes to two UI's and therefore not worth it IMO.

children/schema.py
karisal-anders: We can have a talk about it. IMHO the docstring should reflect what's implemented or at least it shouldn't be misleading i.e. give false information. Now the docstring says "To be considered as being in the past an end time needs to be actually KUKKUU_ENROLLED_OCCURRENCE_IN_PAST_LEEWAY mins in the past." and ..."Event's ticket_system_end_time is in the past". I read those together to mean that past_enough would be correct to be used here. I'm not saying you must use past_enough here though but if you don't use past_enough here then the docstring gives false or at least IMHO misleading information. So please would you either update the docstring to reflect the implementation or the implementation to reflect the docstring, thank you!

tuomas777: I totally agree. I actually forgot to update the docstring when I changed the implementation 😄 Good catch! 👍 Fixed now.

children/schema.py
```diff
@@ -224,17 +224,30 @@ def resolve_active_internal_and_ticket_system_enrolments(self, info, **kwargs):
...
+        datetime_max = make_aware(datetime(9999, 2, 18))
         return sorted(
             (
                 *EnrolmentNode.get_queryset(internal_enrolments, info),
                 *TicketmasterEnrolmentNode.get_queryset(ticket_system_passwords, info),
             ),
-            key=lambda e: e.time,
+            # Sort events without ticket system end time as last ones and then sort
+            # them by published at.
+            key=lambda e: (e.time or datetime_max, e.published_at),
```
karisal-anders, line 250: There's [datetime.max](https://docs.python.org/3/library/datetime.html#datetime.datetime.max) which gives `datetime(MAXYEAR, 12, 31, 23, 59, 59, 999999, tzinfo=None)` but then that might break when changing the timezone as it might go over the maximum. Hmm...

One possibility to remove the datetime_max altogether if it e.time sorting is enough for those that have it:
```python
key=lambda e: (e.time is None, e.time or e.published_at),
```
but if they should also be sorted secondarily by e.published then:
```python
key=lambda e: (e.time is None, e.time or e.published_at, e.published_at),
```
but that gets complicated.

What's the reason for using 9999-02-18? Why not 9999-12-31?

tuomas777: Yeah, `datetime.max` did not work, and neither did 31.12. (date value out of range). 18.2. is just an arbitrary choice, and if that happens to be someone's birthday, it is just a coincidence.

karisal-anders: Hmm... trying it out in django shell:
```python
make_aware(datetime(9999, 12, 31))

/usr/local/lib/python3.9/site-packages/django/utils/timezone.py in make_aware(value, timezone, is_dst)
    240     if _is_pytz_zone(timezone):
    241         # This method is available for pytz time zones.
--> 242         return timezone.localize(value, is_dst=is_dst)
    243     else:
    244         # Check that we won't overwrite the timezone of an aware datetime.

/usr/local/lib/python3.9/site-packages/pytz/tzinfo.py in localize(self, dt, is_dst)
    321         possible_loc_dt = set()
    322         for delta in [timedelta(days=-1), timedelta(days=1)]:
--> 323             loc_dt = dt + delta
    324             idx = max(0, bisect_right(
    325                 self._utc_transition_times, loc_dt) - 1)

OverflowError: date value out of range
```
So pytz's [localize](https://github.com/stub42/pytz/blob/df59d3af429d46ee575ab5d3a7fe3f8cd49a74bc/src/pytz/tzinfo.py#L322) which is called by django's make_aware adds one day to the value which goes over datetime.max. Ok, now I understand.

tuomas777: Changed it to `make_aware(datetime.max, timezone=timezone.utc)`.

nikomakela (approved):
Nice! Good job

PR #303, opened 2022-10-20
KK-973 | Ticketmaster event related fixes

tuomas777:


children/models.py
```diff
@@ -83,16 +83,18 @@ def can_user_administer(self, user):
...
         now = timezone.now()
         year = year or now.year
 
         occurrence_filters = Q(time__year=year)
-        ticket_system_filters = Q(event__occurrences__time__year=year)
+        ticket_system_filters = Q(event__published_at__year=year)
         if past:
             occurrence_filters &= Q(time__lt=now)
-            ticket_system_filters &= Q(event__occurrences__time__lt=now)
+            ticket_system_filters &= Q(event__published_at__lt=now)
```
nikomakela, line 97: Cannot the next years events be published before the year starts?

nikomakela (approved):
code wise this is OK, so I approve this, but I remember several meetings with PO where we discussed about the events that are published e.g in December but are held (at least partly) in January. 

1. For example, if the Ticketmaster event is in the start of January (and the event date tells that), I wouldn't filter it with publish date but the event date
2. If an internally enrollable event has some of the occurrences in December, but some in January, or the event is published in December but is held in January, is the published date the right one?

PR #306, opened 2022-12-01
KK-976 | Improve occurrence queries' performance

tuomas777:
Added some database prefetching and new count fields `attendedEventCount` and `freeSpotNotificationSubscriptionCount` to `OccurrenceNode` so that the admin UI can get the needed occurrences data faster.

### DIFF
--- a/events/models.py
+++ b/events/models.py
@@ -6,7 +6,7 @@ from typing import Optional
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models, transaction
-from django.db.models import ExpressionWrapper, F, Q, UniqueConstraint
+from django.db.models import Count, ExpressionWrapper, F, Q, UniqueConstraint
 from django.db.models.functions import Coalesce
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
@@ -425,6 +425,23 @@ class OccurrenceQueryset(models.QuerySet):
             ),
         )
 
+    def with_enrolment_count(self):
+        return self.annotate(enrolment_count=Count("enrolments", distinct=True))
+
+    def with_attended_enrolment_count(self):
+        return self.annotate(
+            attended_enrolment_count=Count(
+                "enrolments", Q(enrolments__attended=True), distinct=True
+            )
+        )
+
+    def with_free_spot_notification_subscription_count(self):
+        return self.annotate(
+            free_spot_notification_subscription_count=Count(
+                "free_spot_notification_subscriptions", distinct=True
+            )
+        )
+
 
 class Occurrence(TimestampedModel):
     time = models.DateTimeField(verbose_name=_("time"))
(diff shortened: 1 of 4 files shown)


events/models.py
```diff
@@ -518,6 +535,20 @@ def get_enrolment_count(self):
         except AttributeError:
             return self.enrolments.count()
 
+    def get_attended_enrolment_count(self):
+        try:
+            # try to use an annotated value
+            return self.attended_enrolment_count
+        except AttributeError:
+            return self.enrolments.filter(attended=True).count()
```
karisal-anders, line 543: As the annotated value uses distinct, should distinct be added here as well?

events/models.py
```diff
@@ -518,6 +535,20 @@ def get_enrolment_count(self):
...
+
+    def get_free_spot_notification_subscription_count(self):
+        try:
+            # try to use an annotated value
+            return self.free_spot_notification_subscription_count
+        except AttributeError:
+            return self.free_spot_notification_subscriptions.count()
```
karisal-anders, line 550: As the annotated value uses distinct, should distinct be added here as well?

events/models.py
```diff
@@ -518,6 +535,20 @@ def get_enrolment_count(self):
         except AttributeError:
             return self.enrolments.count()
```
karisal-anders, line 536: As the annotated value uses distinct, should distinct be added here as well?

events/tests/test_api.py
```diff
@@ -679,6 +680,51 @@ def test_occurrences_query_project_user(
     snapshot.assert_match(executed)
 
 
+OCCURRENCES_COUNTS_QUERY = """
+query Occurrences {
+  occurrences {
+    edges {
+      node {
+        enrolmentCount
+        %s
```
karisal-anders, line 689: Looking at these *_QUERY variables in the codebase none of the others need interpolation to be used. Just thinking whether it could add clarity to the code to name this one differently, e.g. OCCURENCES_COUNT_QUERY_TEMPLATE or something like that because it won't work without string interpolation. Naming the variable to be interpolated might also add clarity.

events/models.py
karisal-anders: After discussion this should be fine, there shouldn't be need for distinct here because "enrolments" is the related_name in the ForeignKey references from Enrolment to Occurence.

events/models.py
karisal-anders: After discussion this should be fine, there shouldn't be need for distinct here because "enrolments" is the related_name in the ForeignKey references from Enrolment to Occurence and adding the boolean filter shouldn't change things either here AFAIK.

events/models.py
karisal-anders: After discussion this should be fine, there shouldn't be need for distinct here because "free_spot_notification_subscriptions" is the related_name in the ForeignKey references from FreeSpotNotificationSubscription to Occurence.

venues/schema.py
```diff
@@ -82,7 +82,10 @@ def resolve_occurrences(self, info, **kwargs):
         )
 
     def resolve_translations(self, info, **kwargs):
-        return self.translations.order_by("language_code")
+        try:
+            return self.prefetched_translations
+        except AttributeError:
+            return self.translations.order_by("language_code")
```
karisal-anders, line 88: The order_by ascending/descending order differs between the prefetched_translations and this version (i.e. order_by("-language_code") vs. order_by("language_code")).

nikomakela (approved):

events/tests/test_api.py
nikomakela: I would create a query getter out of it

venues/schema.py
tuomas777: Well spotted 👀. Changed.

PR #312, opened 2023-03-03
Process notifications sending outside the atomic transaction of saving of events

nikomakela:
KK-984.
The object instance save-process should be handled in a transaction,
so if any errors occurs, the whole process is rolled back.
The sending of notifications (emails and sms) cannot be in that same transaction,
because a sent message cannot be rolled back --
the receiver has already received the message.

The notifications context should be processed as much as possible
before the actual sending, so that any issues should have been already raised,
before any of the messages has been delivered yet.

### DIFF
--- a/events/tests/test_notifications.py
+++ b/events/tests/test_notifications.py
@@ -1,9 +1,11 @@
 from copy import deepcopy
 from datetime import timedelta
+from unittest import mock
 
 import pytest
 from django.core import mail
 from django.core.management import call_command
+from django.db import Error
 from django.utils.timezone import now
 from freezegun import freeze_time
 from graphql_relay import to_global_id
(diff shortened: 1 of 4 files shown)


nikomakela:
NOTE: The build & review has been failing for a while already. There is a problem in Kolga!

events/tests/test_notifications.py
```diff
@@ -166,6 +168,31 @@ def test_event_publish_notification(
...
+    event_variables = deepcopy(PUBLISH_EVENT_VARIABLES)
+    event_variables["input"]["id"] = to_global_id("EventNode", unpublished_event.id)
+
+    with mock.patch(
+        "events.models.Event.publish",
+        side_effect=Error("Some error occured"),
+    ):
+        publisher_api_client.execute(PUBLISH_EVENT_MUTATION, variables=event_variables)
+        # No mails sent because publication failed
+        assert len(mail.outbox) == 0
```
karisal-anders, line 193: Using mock.patch like this AFAIK results in the original events.models.Event.publish function not being called at all and only the side_effect Error being raised, tried it out locally using pytest and if confirms this behavior. So, this is testing that if events.models.Event.publish would be a function that does nothing else besides raise an Error then no mail is sent? Maybe overriding the events.models.Event.save function and making it fail would be a better test? i.e.
```python
    with mock.patch(
        "events.models.Event.save",
        side_effect=Error("Some error occured"),
    ):
        publisher_api_client.execute(PUBLISH_EVENT_MUTATION, variables=event_variables)
        # No mails sent because publication failed
        assert len(mail.outbox) == 0
```
What do you think?

events/utils.py
```diff
@@ -36,14 +41,31 @@ def send_event_notifications_to_guardians(
...
+                context_for_address[guardian.email] = context
 
-                send_notification(
-                    guardian.email,
-                    notification_type,
-                    context=context,
-                    language=guardian.language,
-                    attachments=attachments,
-                )
+    emails = ",".join([address for address in context_for_address.keys()])
```
karisal-anders, line 46:
```python
emails = ",".join(context_for_address.keys())
```
should work fine, tried this:
```python
>>> ','.join({"a":1, "b":2, "c":3})
'a,b,c'
```

events/utils.py
```diff
@@ -36,14 +41,31 @@ def send_event_notifications_to_guardians(
...
+    logger.debug(
+        f"There are ${len(emails)} total in the notification list. "
```
karisal-anders, line 48: The `$` character is printed as is here

events/tests/test_notifications.py
nikomakela: Nice catch! Changed.

nikomakela:
> This fixes [KK-984](https://helsinkisolutionoffice.atlassian.net/browse/KK-984) and AFAIK a possibility that all events are marked as published but no SMS or email notifications are sent at all remains, which is not KK-984 but in a sense the opposite of KK-984.

Yea, most likely, if the notification sending was the issue, it will still be an issue, but now we get more logs related to the notification sendings and at least the event or the event group should be set as published.

PR #313, opened 2023-03-09
Add Lippupiste as an external ticket system

nikomakela:
KK-987 KK-994.

The integration for the external ticket systems,
the Ticketmaster and the Lippupiste, should be similar.
Only thing that varies at this point , is the ticket system URL.

The old Ticketmaster model and schema classes are now renamed
and converted to some abstract classes. The old Ticketmaster class names
are now given to the new classes that inherits the abstracts.
New Lippupiste classes are copied from the new Ticketmaster ones.

The old implementation for Ticketmaster should be unchanged,
so all the existing unit tests should pass without any changes.
There are now some changes added there only to randomly test that also
the Lippupiste events are calculated in same manners.

### DIFF
--- a/children/schema.py
+++ b/children/schema.py
@@ -135,7 +135,7 @@ class ChildNode(DjangoObjectType):
           * external ticket system Events the user has a password to and the Event's
             ticket_system_end_time is in the past at all.
         """
-        from events.schema import OccurrenceNode, TicketmasterEnrolmentNode  # noqa
+        from events.schema import ExternalTicketSystemEnrolmentNode, OccurrenceNode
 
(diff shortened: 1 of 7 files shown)


karisal-anders (commented):
Why the three added and commented out `get_queryset` functions?

nikomakela:
**Let's wait for product owner approval before merging this!** 🚫

HeikkM (approved):

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr313.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #332, opened 2023-12-13
Fix the children API query pagination and offset

nikomakela:
KK-1049

The offset was handled twice, because it was handled 1 once in the custom implementation and once in the super class implementation. That lead to a situation where the pagination was broken, because the offset-parameter is always used to populate the after-parameter and now the offset-parameter was deducted twice, so once too many. Since the super-class implementation already does the work, the offset handling could be removed from the custom implementation.

The related snapshots are now updated and double checked that they are now right. Note that they weren't earlier! Also all the newly added unit tests to test the Children query's cursor generation were added there before the actual fix was implemented. They were failing before the fix, but now they do pass.

There were also some inconsistencies in ordering the result set of the unit tests. Some default orderings were added -- hopefully they fix the issues.


children/tests/snapshots/snap_test_api.py
```diff
@@ -154,30 +154,42 @@
...
                 {"node": {"lastName": "3"}},
                 {"node": {"lastName": "4"}},
+                {"node": {"lastName": "5"}},
             ]
         }
     }
 }
 
 snapshots["test_children_offset_pagination[2-2] 1"] = {
-    "data": {"children": {"edges": [{"node": {"lastName": "4"}}]}}
```
nikomakela, line 168: This was certainly not right, because the limit was set to 2 and only 1 item was returned!

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr332.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr332.api.dev.hel.ninja :laughing::tada::tada::tada:

children/tests/test_api.py
```diff
@@ -1235,6 +1242,148 @@ def test_children_offset_pagination(
...
+    )
+
+
+@pytest.mark.parametrize("page", list(range(1, 5)))
+@pytest.mark.parametrize("limit", [1, 2, 10, 20])
+def test_children_pagination_cursor_generation(
+    project_user_api_client, project, page, limit
+):
+    """
+    Test that the cursors are created as invented.
```
karisal-anders, line 1347: What do you mean by "as invented"? Something similar to "as expected" or something else? Maybe the wording could be made clearer here although the intent can be guessed.

karisal-anders (approved):
Added a comment about an unclear wording, otherwise LGTM to merge

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr332.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr332.api.dev.hel.ninja :laughing::tada::tada::tada:

PR #336, opened 2024-01-19
Simplify the name fields of the Child model

nikomakela:
KK-1024 KK-1078.

1. Remove the last name field from the Child model.
2. Rename the  first name field of the child model to "name".

There is no need for actual first name or last name in Child domain.
Something like a nick name or a calling name or calling sign
should be just enough, because the name field is there only
to identify the object instance.


karisal-anders (approved):
Looks otherwise ok but not sure whether renaming the `first_name` field to `name` field is really a good thing or not. Sure we could in the UI rename it again or present it as something else like `calling name` or `nickname` to make people not put in the whole name of the child, but wouldn't it be better have the intention of not having a whole name in this field in the model itself already? I mean, perhaps just leaving it as `first_name` would be better to give that intent? Or renaming to `calling_name` or `nickname`, but that's a bit more moot.

The point being that if we want to make sure the field is used for less GDPR sensitive data (i.e. only first name or calling name or nickname rather than the whole name) then preserving that intention in the field name also would IMHO make sense.

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr336.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr336.api.dev.hel.ninja :laughing::tada::tada::tada:

PR #337, opened 2024-01-19
Migrate Child birthdate to be birthyear

nikomakela:
KK-1024 KK-1077.

The date of the birth is not needed, but the year is. The birthdate field is first renamed to be birthyear. After field name change, the field type is converted from a date to an integer. Data migration is needed to be ran first, so that we don't lose the birth year that is already filled within the date.


karisal-anders (approved):
Otherwise looks good but it would be good to be consistent with `birth year` vs. `birthyear`. There's now two different ways it is being used in the code (`birth_year` and `birthyear`), and looking online it seemed to me that `birth year` is more common than `birthyear`.

https://www.google.com/search?q=%22birthyear%22 -> 676k results
https://www.google.com/search?q=%22birth%20year%22 -> 14,7M results

Did you check that the data migration using the "USING EXTRACT..." worked correctly?

nikomakela:
> Otherwise looks good but it would be good to be consistent with `birth year` vs. `birthyear`. There's now two different ways it is being used in the code (`birth_year` and `birthyear`), and looking online it seemed to me that `birth year` is more common than `birthyear`.
> 
> https://www.google.com/search?q=%22birthyear%22 -> 676k results https://www.google.com/search?q=%22birth%20year%22 -> 14,7M results
> 
> Did you check that the data migration using the "USING EXTRACT..." worked correctly?

I first used the `birth_year`there too, but then when I started replacing the values in Graphql schema code, I changed the field name back to the same structure as it was given earlier -- `birthdate` -> `birthyear`. The schema uses low-camel-case formatting on field names, so it would become `birthYear`. Now when they all are `birthyear`, it was just easier for the coders who maintain the code and the structure/Formatting was approved earlier too.

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr337.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr337.api.dev.hel.ninja :laughing::tada::tada::tada:

PR #360, opened 2024-05-03
Add obsoleted props to User and Child

nikomakela:
KK-1131.

1. The User model now has a field is_obsoleted to indicate whether the User instance is still accessible.
2. The child is considered obsoleted if the child does not have any guardians or all the guardian users linked to it are obsoleted.

NOTE: There should not be any children with multiple guardians in the database, because the UI has never allowed it, but the database schema allows it.

The following features are added:
- obsoleted queryset for Child
- filter reports of children by is_obsolete
- add the user is_obsolete field to child report serializer
- refactor user admin so that it gives more value
- add the is_obsolete to the user admin
- add is_obsolete field to the user model


terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr360.api.dev.hel.ninja** :rocket::rocket::rocket:

reports/api.py
```diff
@@ -178,6 +186,13 @@ class ChildViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
             ),
         )
         .prefetch_related("guardians")
+        .prefetch_related("guardians__user")
```
nikomakela, line 189: I wonder whether this would be as good with just a `prefetch_related("guardians__user")`. Could the `prefetch_related("guardians")` be redundant then?

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr360.api.dev.hel.ninja :laughing::tada::tada::tada:

reports/api.py
nikomakela: It seems to not create anymore queries, because there is a `django_assert_max_num_queries` tester in the `reports/tests/test_api.py::test_children_endpoint`.

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr360.api.dev.hel.ninja :laughing::tada::tada::tada:

children/models.py
```diff
@@ -27,6 +27,24 @@ def user_can_update(self, user):
...
         return self.filter(guardians__user=user)
 
+    def obsoleted(self, is_obsolete: bool = True):
+        """The child is considered obsoleted
+        if the child does not have any guardians or
+        all the guardian users linked to it are obsoleted.
+
+        Args:
+            is_obsolete (bool, optional): filter by the obsolete
+            or non-obosolete children. Defaults to True.
```
karisal-anders, line 37: typo: `obosolete` -> `obsolete`

children/models.py
```diff
@@ -153,6 +171,25 @@ def name_birthyear_postal_code_guardian_emails_hash(self) -> str:
             [self.name] + self._salt_birthyear_postal_code_guardian_emails_list
         )
 
+    @property
+    def is_obsolete(self) -> bool:
+        """The child is obsoleted if they doesn't have any guardians or
```
karisal-anders, line 176: they doesn't -> they don't

children/models.py
```diff
@@ -153,6 +171,25 @@ def name_birthyear_postal_code_guardian_emails_hash(self) -> str:
...
+        (when the Tunnistamo is changed to a Keycloak service).
+
+        Returns:
+            bool: returns True, if there are no guardians or
+            all the users are marked as obsoleted,
+            otherwise False.
+        """
+        return self.guardians.count() == 0 or all(
+            [guardian.user.is_obsolete for guardian in self.guardians.all()]
+        )
```
karisal-anders, line 191: In this context a micro-optimization and a non-issue but the list comprehension isn't needed i.e.
```python
return self.guardians.count() == 0 or all(
            guardian.user.is_obsolete for guardian in self.guardians.all()
        )
```
will work using the generator function `guardian.user.is_obsolete for guardian in self.guardians.all()` directly.

reports/api.py
```diff
@@ -110,6 +111,9 @@ class ChildSerializer(serializers.ModelSerializer):
         help_text="Array of ISO 639-3 (language) or ISO 639-5 (language family) "
         "alpha-3 codes. Value `__OTHER__` means any other language."  # noqa
     )
+    is_obsolete = serializers.SerializerMethodField(
+        help_text="Is all the child's guardian users instances marked as obsolete?"
```
karisal-anders, line 115: How about
"Are all the child's guardian user instances marked as obsolete?" or
"Are all the child's guardian users' instances marked as obsolete?"

reports/api.py
```diff
@@ -178,6 +186,13 @@ class ChildViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
...
+        .prefetch_related("guardians__user")
         .order_by("name")
     )
     serializer_class = ChildSerializer
+
+    def get_queryset(self):
+        filters = {}
+        if is_obsolete := self.request.query_params.get("is_obsolete"):
+            filters["guardians__user__is_obsolete"] = bool(strtobool(is_obsolete))
+        return super().get_queryset().filter(**filters)
```
karisal-anders, line 198: This will throw a ValueError if is_obsolete query parameter is not valid, e.g. 'maybe':
```python
def strtobool (val):
    """Convert a string representation of truth to true (1) or false (0).

    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
    'val' is anything else.
    """
    val = val.lower()
    if val in ('y', 'yes', 't', 'true', 'on', '1'):
        return 1
    elif val in ('n', 'no', 'f', 'false', 'off', '0'):
        return 0
    else:
        raise ValueError("invalid truth value %r" % (val,))
```
which will result in internal server error probably? Is that acceptable or should the invalid values rather fall back to not taking the parameter into account at all? How do you see this?

karisal-anders: BTW [distutils is deprecated](https://peps.python.org/pep-0632/) and was removed in Python 3.12.

karisal-anders (approved):
Left some comments, please take a look at them, the use of the deprecated distutils package's strtobool (removed in Python 3.12) might be good to consider changing to something else IMHO. Overall looks quite ok.

children/tests/test_models.py
```diff
@@ -190,3 +190,74 @@ def test_child_serialize(snapshot, project):
...
+    queryset = Child.objects.obsoleted(is_obsolete=False)
+    # then the children of the non-obsoleted guardian user
+    # should be returned.
+    assert queryset.count() == 1
+    assert all(not child.is_obsolete for child in queryset)
+
+
+@pytest.mark.django_db
+def test_child_obsoleted_queryset_when_not_all_the_guardians_are_obsoleted():
+    """Test that the is_obsoleted queryset works correctly
```
karisal-anders, line 247: `is_obsoleted` -> `obsoleted`

reports/api.py
nikomakela: I copied this function and used try-except

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr360.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #361, opened 2024-05-03
feat: add is_obsoleted variable to the relevant email templates

nikomakela:
KK-1131


karisal-anders (commented):
In PR #360 you used `is_obsolete` and `obsoleted`, so IMHO it'd be better to use either of those terms here also and not a third combination.

karisal-anders (approved):
Otherwise ok, but IMHO it'd be better to use either of the terms already used in PR #360 i.e. `is_obsolete` or `obsoleted` instead of coming up with a third alternative `is_obsoleted`.

PR #363, opened 2024-05-24
feat: send event and event group publish notification in a thread

nikomakela:
KK-1035.

Added a new common decorator that can be used to execute tasks in background
in a new thread.

If there are thousands of email recipients, the process of creating the mails
will take tens of seconds. While the process is handled in the same
request handler as the event state change, the response takes long to be sent,
which leads to a connection timeout, that will write an error message
to the admin UI. To prevent that, the state change could be done as an atomic
operation as it has always been done, but the emails could be sent in an another
thread. The response is then sent when the atomic process finishes and the email
sending is not awaited.

More about the threading for email sending:
- https://ozcanyarimdunya.github.io/dj_threading/
- https://medium.com/@akshatgadodia/simplifying-email-sending-in-django-a-step-by-step-guide-7dc90079738b
fix: db connection was not working properly in pytests

Some of the pytests needed some threading handling: the database connection was not shared 
with the newly spawned threads, so e.g. the sent mails were not shown in the mail box because
the querysets were empty. Using `@pytest.mark.django_db(transaction=True)` helped in that.

----

NOTE: The threading might be a problem with our UWSGI!

"Django threading": https://ozcanyarimdunya.github.io/dj_threading/ 

"Sending Emails in the Background with Multithreading": https://medium.com/@akshatgadodia/simplifying-email-sending-in-django-a-step-by-step-guide-7dc90079738b. 

"Django-allauth using threads": https://github.com/pennersr/django-allauth/issues/1451. 

"Threads should be used", "Celery is definitely an Overkill for just sending emails..." https://stackoverflow.com/questions/4447081/how-to-send-asynchronous-email-using-django. 

"Using threads for offloading request is not so good idea...": https://stackoverflow.com/questions/17507968/django-threading-and-tests?rq=4 

"Threads are bad solution...": https://stackoverflow.com/questions/13567160/django-best-way-to-send-email-in-background. 

"Process or Thread problem deploying Django app with uWSGI": https://forum.djangoproject.com/t/process-or-thread-problem-deploying-django-app-with-uwsgi/22189 

UWSGI configuration for threading: https://stackoverflow.com/questions/58389626/multiprocessing-process-is-blocking-the-return-statement-when-running-with-uwsgi.

### DIFF
--- a/common/decorators.py
+++ b/common/decorators.py
@@ -1,0 +1,37 @@
+import logging
+from functools import wraps
+from threading import Thread
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+
+def execute_in_background(
+    thread_name: Optional[str] = None, daemonic: Optional[bool] = None
+):
+    """A decorator to execute a function in a new thread in background.
+
+    Args:
+        daemonic (Optional[bool], optional): Set a new thread as daemonic.
+            If not None, daemon explicitly sets whether the thread is daemonic.
+            If None, the daemonic property is inherited from the current thread.
+            Defaults to None.
+    """
+
+    def decorator(target_function):
+        @wraps(target_function)
+        def start_thread(*args, **kwargs):
+            logger.debug("Creating a new thread to execute a function in background...")
+            thread = Thread(
+                name=thread_name,
+                target=target_function,
+                args=args,
+                kwargs=kwargs,
+                daemon=daemonic,
+            )
+            logger.info(f"Starting a thread [{thread}].")
+            thread.start()
+
+        return start_thread
+
+    return decorator
--- a/events/tests/test_notifications.py
+++ b/events/tests/test_notifications.py
@@ -1,3 +1,4 @@
+import threading
 from copy import deepcopy
 from datetime import timedelta
 from unittest import mock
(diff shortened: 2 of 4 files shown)


terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr363.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr363.api.dev.hel.ninja :laughing::tada::tada::tada:

nikomakela:
To test, some sleep can be added for example to the events/utils.py notification sending commands.

```
    import time

    time.sleep(10)
    print("sleeping for 10")
```

common/decorators.py
```diff
@@ -0,0 +1,37 @@
...
+        def start_thread(*args, **kwargs):
+            logger.debug("Creating a new thread to execute a function in background...")
+            thread = Thread(
+                name=thread_name,
+                target=target_function,
+                args=args,
+                kwargs=kwargs,
+                daemon=daemonic,
+            )
+            logger.info(f"Starting a thread [{thread})].")
```
karisal-anders, line 32: An extra `)` here

events/tests/test_notifications.py
```diff
@@ -33,6 +34,19 @@
 from users.factories import GuardianFactory
 
 
+def _join_thread_with_name(thread_name: str, ignore_errors=False):
```
karisal-anders, line 37: Maybe the name could be made clearer/easier to understand at a glance? For example `_wait_for_thread` or something similar?

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr363.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr363.api.dev.hel.ninja :laughing::tada::tada::tada:

PR #374, opened 2024-06-10
Make it possible to send a mass email about the auth service is changing

nikomakela:
KK-1138 KK-1140 KK-1141 KK-1142.

Refactored the old notification service code a bit by moving the functions from utils-file to new classes in services.py.

The notification template should offer the default dates for the service change, because otherwise it would always need a developer to generate a new notification content.

A management command that has the following inputs:
- date of the change of the auth service can be given as a formatted string, e.g. "17.6.2024". Use it with "-d 17.6.2024".
- a user joined date filter can be used to filter the queryset of the guardians. Use "-j 2024-06-06T15:00:00".
- by default the queryset includes only the obsoleted guardians, but all the guardians can be included with "--include_non_obsoleted".
- a list of emails can be used as a recipient filter. Use -e.

In the guardians admin list view, you can select one guardian and generate
1. the markdown of the guardian's children history of events.
2. full notification template.
Screenshot 2024-06-10 at 18 38 00

An example of the generated markdown can be seen from the unit test snapshots.

### DIFF
--- a/users/admin.py
+++ b/users/admin.py
@@ -5,7 +5,10 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import GroupAdmin as DjangoGroupAdmin
 from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
 from django.contrib.auth.models import Group
+from django.http import HttpResponse
 from django.utils.translation import gettext_lazy as _
+from django_ilmoitin.models import NotificationTemplate
+from django_ilmoitin.utils import render_notification_template
 from guardian.admin import GuardedModelAdmin
 
 from children.models import Relationship
@@ -13,6 +16,8 @@ from languages.models import Language
 from projects.models import Project
 from reports.models import Permission as ReportPermission
 from users.models import Guardian
+from users.notifications import NotificationType
+from users.services import AuthServiceNotificationService
 
 
 class RelationshipInline(admin.TabularInline):
@@ -36,6 +41,115 @@ class LanguagesSpokenAtHomeInline(admin.TabularInline):
     extra = 0
     verbose_name = _("Language spoken at home")
     verbose_name_plural = _("Languages spoken at home")
+
+
+def _generate_children_event_history_markdown(modeladmin, request, queryset):
+    if queryset.count() != 1:
+        modeladmin.message_user(
+            request,
+            "Only 1 guardian can be selected for "
+            "'Generate authorization server is changing notification text' action.",
+        )
+        return
+    guardian = queryset.first()
+    return (
+        guardian,
+        AuthServiceNotificationService.generate_children_event_history_markdown(
+            guardian=guardian
+        ).replace("\n", "<br/>"),
+    )
+
+
+@admin.action(description="Generate children event history markdown")
+def generate_children_event_history_markdown(modeladmin, request, queryset):
+    (
+        guardian,
+        children_event_history_markdown,
+    ) = _generate_children_event_history_markdown(modeladmin, request, queryset)
+    return HttpResponse(children_event_history_markdown)
+
+
+def _generate_generate_user_auth_service_is_changing_notification_for_lang(
+    language: str,
+):
+    def _generate_user_auth_service_is_changing_notification_text(
+        modeladmin, request, queryset
+    ):
+        (
+            guardian,
+            children_event_history_markdown,
+        ) = _generate_children_event_history_markdown(modeladmin, request, queryset)
+
+        template = NotificationTemplate.objects.filter(
+            type=NotificationType.USER_AUTH_SERVICE_IS_CHANGING
+        ).first()
+
+        if not template:
+            modeladmin.message_user(
+                request,
+                "USER_AUTH_SERVICE_IS_CHANGING is not yet available in the database. "
+                "It should be importable from the notifications spreadsheet?",
+            )
+            return
+
+        context = {
+            "guardian": guardian,
+            "date_of_change_str": None,  # give default in notification template instead
+            "children_event_history_markdown": children_event_history_markdown,
+        }
+
+        subject, body_html, body_text = render_notification_template(
+            template, context, language
+        )
+
+        return f"""
+    {subject}
+
+    {body_text}
+    """.replace(
+            "\n", "<br/>"
+        )
+
+    return _generate_user_auth_service_is_changing_notification_text
+
+
+@admin.action(
+    description="Generate children event history notification email content (fi)"
+)
+def generate_user_auth_service_is_changing_notification_text_fi(
+    modeladmin, request, queryset
+):
+    return HttpResponse(
+        _generate_generate_user_auth_service_is_changing_notification_for_lang("fi")(
+            modeladmin, request, queryset
+        )
+    )
+
+
+@admin.action(
+    description="Generate children event history notification email content (sv)"
+)
+def generate_user_auth_service_is_changing_notification_text_sv(
+    modeladmin, request, queryset
+):
+    return HttpResponse(
+        _generate_generate_user_auth_service_is_changing_notification_for_lang("sv")(
+            modeladmin, request, queryset
+        )
+    )
+
+
+@admin.action(
+    description="Generate children event history notification email content (en)"
+)
+def generate_user_auth_service_is_changing_notification_text_en(
+    modeladmin, request, queryset
+):
+    return HttpResponse(
+        _generate_generate_user_auth_service_is_changing_notification_for_lang("en")(
+            modeladmin, request, queryset
+        )
+    )
 
 
 @admin.register(Guardian)
@@ -62,6 +176,12 @@ class GuardianAdmin(admin.ModelAdmin):
     form = GuardianForm
     inlines = (RelationshipInline, LanguagesSpokenAtHomeInline)
     list_filter = ("children__project", "has_accepted_communication")
+    actions = (
+        generate_children_event_history_markdown,
+        generate_user_auth_service_is_changing_notification_text_fi,
+        generate_user_auth_service_is_changing_notification_text_sv,
+        generate_user_auth_service_is_changing_notification_text_en,
+    )
 
 
 class PermissionFilterMixin:
--- a/users/admin.py
+++ b/users/admin.py
@@ -1,3 +1,5 @@
+from typing import Tuple
+
 from django import forms
 from django.conf import settings
 from django.contrib import admin
@@ -18,6 +20,12 @@ from reports.models import Permission as ReportPermission
 from users.models import Guardian
 from users.notifications import NotificationType
 from users.services import AuthServiceNotificationService
+
+USER_AUTH_SERVICE_IS_CHANGING_NOTIFICATION_TEMPLATE = """
+{subject}
+
+{body_text}
+""".strip()
 
 
 class RelationshipInline(admin.TabularInline):
@@ -43,7 +51,9 @@ class LanguagesSpokenAtHomeInline(admin.TabularInline):
     verbose_name_plural = _("Languages spoken at home")
 
 
-def _generate_children_event_history_markdown(modeladmin, request, queryset):
+def _generate_children_event_history_markdown(
+    modeladmin, request, queryset
+) -> Tuple[Guardian, str]:
     if queryset.count() != 1:
         modeladmin.message_user(
             request,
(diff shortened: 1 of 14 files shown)


users/services.py
```diff
@@ -0,0 +1,53 @@
+from django_ilmoitin.utils import send_notification
+
+from users.models import Guardian
+from users.notifications import NotificationType
+from users.utils import get_communication_unsubscribe_ui_url
+
+
+class GuardianEmailManagementNotificationService:
```
karisal-anders, line 11:
```suggestion
class GuardianEmailNotificationService:
```
The class name already tells it's about emails and notifications so
IMHO removing `Management` would be beneficial as it'd make the
class name a bit shorter. How do you see this?

nikomakela:
The unit tests does not run in pipeline (but runs locally). They need to be fixed.

- [ ] Fix CI unit tests

karisal-anders:
@nikomakela "Static tests Sonarcloud with pytest" is failing i.e. https://dev.azure.com/City-of-Helsinki/kukkuu/_build/results?buildId=179284&view=logs&j=77f1ce95-332e-5490-6b22-9d79e3e59f5c&t=f260fe5f-6a98-5f6f-8ad2-564e4fae7b57
> django.db.utils.ProgrammingError: relation "children_child" does not exist

karisal-anders:
> The unit tests does not run in pipeline (but runs locally).

`docker compose up --build` fails for me locally with error `django.db.utils.ProgrammingError: relation "children_child" does not exist`

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr374.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr374.api.dev.hel.ninja :laughing::tada::tada::tada:

users/services.py
karisal-anders: Ah, so how about `GuardianEmailChangeNotificationService` to make it explicit what's it for or adding some documentation to the class to indicate its purpose?

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr374.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr374.api.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr374.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr374.api.dev.hel.ninja :laughing::tada::tada::tada:

users/admin.py
```diff
@@ -38,6 +43,115 @@ class LanguagesSpokenAtHomeInline(admin.TabularInline):
...
+    return (
+        guardian,
+        AuthServiceNotificationService.generate_children_event_history_markdown(
+            guardian=guardian
+        ).replace("\n", "<br/>"),
+    )
+
+
+@admin.action(description="Generate children event history markdown")
+def generate_children_event_history_markdown(modeladmin, request, queryset):
```
karisal-anders, line 64: Hmm... I was about to say [PEP 8](https://peps.python.org/pep-0008/) and `modeladmin` → `model_admin` and `queryset` → `query_set` because the classes are [ModelAdmin](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin) and [QuerySet](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.query.QuerySet) and those parameters are probably instances of those classes. But to my surprise Django's [documentation](https://docs.djangoproject.com/en/3.2/ref/contrib/admin/actions/#writing-action-functions) uses this naming convention in its example:
```python
def make_published(modeladmin, request, queryset):
    queryset.update(status='p')
```

So this is acceptable given that Django is using this in its own documentation and PEP 8's points:

1. [Naming conventions](https://peps.python.org/pep-0008/#naming-conventions):
   - > New modules and packages (including third party frameworks) should be written to these standards, but where an existing library has a different style, internal consistency is preferred.
2. [Function and method arguments](https://peps.python.org/pep-0008/#function-and-method-arguments) section in PEP 8 doesn't actually specify naming convention which I find a bit peculiar

users/admin.py
```diff
@@ -38,6 +43,115 @@ class LanguagesSpokenAtHomeInline(admin.TabularInline):
     verbose_name_plural = _("Languages spoken at home")
 
 
+def _generate_children_event_history_markdown(modeladmin, request, queryset):
```
karisal-anders, line 46:
```suggestion
def _generate_children_event_history_markdown(
    modeladmin, request, queryset
) -> Tuple[Guardian, str]:
```
With `from typing import Tuple` for added clarity/documentation about return type.

users/admin.py
```diff
@@ -38,6 +43,115 @@ class LanguagesSpokenAtHomeInline(admin.TabularInline):
...
+    description="Generate children event history notification email content (en)"
+)
+def generate_user_auth_service_is_changing_notification_text_en(
+    modeladmin, request, queryset
+):
+    return HttpResponse(
+        _generate_generate_user_auth_service_is_changing_notification_for_lang("en")(
+            modeladmin, request, queryset
+        )
+    )
```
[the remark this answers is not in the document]
karisal-anders: About the triple-quoted template string, now it contains leading newline and lines have leading spaces. For example in python shell:
```
>>> subject='subject'
>>> body_text='body_text'
>>> x = f"""
...     {subject}
...     {body_text}
...     """.replace("\n", "<br/>")
>>> x
'<br/>    subject<br/>    body_text<br/>    '
```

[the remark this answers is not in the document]
karisal-anders:
```python
return f"{subject}<br/><br/>{body_text}<br/>"
```
If you don't need the indentations nor the leading newline that should probably do it.

users/management/commands/send_user_auth_service_is_changing_notifications.py
```diff
@@ -0,0 +1,81 @@
...
+from django.utils.dateparse import parse_datetime
+
+from users.services import AuthServiceNotificationService
+
+
+class Command(BaseCommand):
+    help = "Sends user authorization service is changing notifications."
+
+    def add_arguments(self, parser):
+        # user_joined_before: Optional[datetime] = None, obsoleted_users_only=True
```
karisal-anders, line 12: It is unclear IMHO what this comment is about. Is this the default behavior or something else?

users/management/commands/send_user_auth_service_is_changing_notifications.py
```diff
@@ -0,0 +1,81 @@
...
+            help="Date of the change of the auth service as a formatted str "
+            "(e.g., '17.6.2024'). "
+            "The default should be given in the notification template.",
+            default=None,  # The default should be given in the notification template.
+        )
+        parser.add_argument(
+            "-j",
+            "--joined_before",
+            type=str,
+            help="Users who has joined before this datetime are included in queryset "
```
karisal-anders, line 26:
```suggestion
            help="Users who have joined before this datetime are included in queryset "
```

users/management/commands/send_user_auth_service_is_changing_notifications.py
```diff
@@ -0,0 +1,81 @@
...
+            "-e",
+            "--emails",
+            nargs="+",  # Accept multiple values as a list
+            help="List of guardian emails to send notifications to (space-separated).",
+        )
+        parser.add_argument(
+            "--include_non_obsoleted",
+            action="store_true",
+            help="Date of the change in auth server. "
+            "The default should be given in the notification template.",
```
karisal-anders, line 40: Help text out-of-sync with the argument

users/services.py
```diff
@@ -0,0 +1,136 @@
...
+            language=guardian.language,
+        )
+
+    @staticmethod
+    def send_user_auth_service_is_changing_notifications(
+        guardians: Optional[Union[QuerySet, list[Guardian]]] = None,
+        date_of_change_str: Optional[str] = None,
+    ):
+        """Send user authentication service is changing notifications
+        to guariands as recipients.
```
karisal-anders, line 99:
```suggestion
        to guardians as recipients.
```

users/services.py
```diff
@@ -0,0 +1,136 @@
...
+        """Generates a Markdown string listing a guardian's children's enrolments."""
+
+        # NOTE: This is markdown, so the line changes and white spaces are important!
+        template_string = """
+{% for child in guardian.children.all %}# {{ child.name }}
+{% for enrolment in child.enrolments.all|dictsort:"occurrence.time" %}1. **{{ enrolment.occurrence.event.name }}:** {{ enrolment.occurrence.time|date:"j.n.Y H:i" }}{% if enrolment.occurrence.event.short_description %}
+{{enrolment.occurrence.event.short_description}}{% endif %}
+{% endfor %}
+{% endfor %}
+"""  # noqa
```
karisal-anders, line 132: How about moving the template string outside of the class so there's no indentation so it reads clearer in relation to its surroundings' indentation level & using `.strip()` to get rid of the leading and trailing newlines unless they're necessary?
i.e.
```python
# NOTE: This is markdown, so the line changes and white spaces are important!
CHILDREN_EVENT_HISTORY_MARKDOWN_TEMPLATE = """
{% for child in guardian.children.all %}# {{ child.name }}
{% for enrolment in child.enrolments.all|dictsort:"occurrence.time" %}1. **{{ enrolment.occurrence.event.name }}:** {{ enrolment.occurrence.time|date:"j.n.Y H:i" }}{% if enrolment.occurrence.event.short_description %}
{{enrolment.occurrence.event.short_description}}{% endif %}
{% endfor %}
{% endfor %}
""".strip()  # noqa
```
How would you feel about this change?

users/tests/test_notifications.py
```diff
@@ -39,6 +43,35 @@ def notification_template_guardian_email_change_requested_fi():
...
+        Occurrence: {{enrolment.occurrence.time}}
+
+        {% endfor %}
+        {% endfor %}
+        {% endif %}
+        {% if children_event_history_markdown %}
+        Markdown: 
+        {{children_event_history_markdown}}
+        {% endif %}
+        """,  # noqa
```
karisal-anders, line 71: There is leading whitespace on all other lines except the first line, is that intentional? How about moving this to a variable/constant just before the pytest.fixture definition to keep it local but still make the indentation problems go away? How would you feel about this change?

users/management/commands/send_user_auth_service_is_changing_notifications.py
```diff
@@ -0,0 +1,81 @@
...
+            default=False,
+        )
+
+    def _get_joined_before(self, **options):
+        joined_before_datetime_str = options["joined_before"]
+        invalid_date_error_str = "Invalid datetime format with joined_before. Use ISO format (YYYY-MM-DDTHH:MM:SS)."  # noqa
```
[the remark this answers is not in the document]
nikomakela: could not get it to work earlier. I don't know why. But I can try again. I tried it exactly like that.

users/tests/test_models.py
```diff
@@ -251,8 +253,12 @@ def test_user_serialize(snapshot, project):
         child_with_one_enrolment,
         child_without_enrolments,
     ] = ChildWithGuardianFactory.create_batch(3, relationship__guardian=guardian)
-    FreeSpotNotificationSubscriptionFactory(child=child_without_enrolments)
-    FreeSpotNotificationSubscriptionFactory(child=child_with_one_enrolment)
+    FreeSpotNotificationSubscriptionFactory(
+        child=child_without_enrolments, occurrence__id=101
+    )
+    FreeSpotNotificationSubscriptionFactory(
+        child=child_with_one_enrolment, occurrence__id=102
```
karisal-anders, line 260: What happens here? Can you override the ID when saving to the database like this or does the database generate the ID value and return it nonetheless if you try to override it? If you need a specific occurrence then why not create it first and reference it? Why the direct ID usage?

karisal-anders (commented):
Left some comments, please take a look at them.

users/admin.py
nikomakela: I won't change the parameter names here since those are decided by the Django framework (or any other 3rd party whose functions we are extending here). In my opinion it would be worse if the variables are renamed by us. Foe example the "queryset" is used all around the django ecosystem.

users/tests/test_models.py
nikomakela: The ID is there to fix the snapshot testing. The ID would change if the tests are run in a different order or only a subset would be ran.

users/services.py
nikomakela: I can change this, but I personally feel that the Management was much more professional

users/management/commands/send_user_auth_service_is_changing_notifications.py
nikomakela: Accident. I'll remove it.

users/management/commands/send_user_auth_service_is_changing_notifications.py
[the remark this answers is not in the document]
nikomakela: ...maybe I missed the brackets.

users/services.py
nikomakela: I also replaced the `1.` with `{{ forloop.counter }}.`, because then it looks better in the email. In the parsed markdown it does not make any change.

nikomakela: MOving the strip would actually change the result a bit: The strip after the TEMPLATE would only strip the template, not the HTML that is rendered.

users/tests/test_notifications.py
karisal-anders: In a test it's not so important whether there's some extra whitespace there or not, so ok, just wanted to bring this up as it's easy to get more whitespace than one wants with the triple-quoted strings when inside deeper indentation levels.

users/services.py
karisal-anders: The strip just strips the leading and trailing whitespace from the triple-quoted string, I mean as it starts with a newline i.e.
in Python shell:
```
>>> """
... x
... """
'\nx\n'
>>> """
... x
... """.strip()
'x'
```

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr374.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr374.api.dev.hel.ninja :laughing::tada::tada::tada:

karisal-anders (approved):

PR #383, opened 2024-06-13
KK-1185 | feat: limit sent emails in auth service change notifications command

karisal-anders:
### feat: limit sent emails in auth service change notifications command

-m / --max_guardian_count can be used with
send_user_auth_service_is_changing_notifications management
command to limit the maximum number of guardians/emails
to be handled at once

refs KK-1185

## Closes

[KK-1185](https://helsinkisolutionoffice.atlassian.net/browse/KK-1185)

[KK-1185]: https://helsinkisolutionoffice.atlassian.net/browse/KK-1185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/users/management/commands/send_user_auth_service_is_changing_notifications.py
+++ b/users/management/commands/send_user_auth_service_is_changing_notifications.py
@@ -39,6 +39,14 @@ class Command(BaseCommand):
             "By default only the obsoleted accounts are included in the query set.",
             default=False,
         )
+        parser.add_argument(
+            "-m",
+            "--max_guardian_count",
+            type=int,
+            help="Maximum number of guardians to send notifications to. "
+            "By default there's no limit.",
+            default=None,
+        )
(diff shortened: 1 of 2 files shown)


nikomakela (commented):
I reviewed the current changes and they look good to me! It just that the current implementation (only) does not make sense without some status field manipulation. @karisal-anders you said you had another ticket for status manipulation, so I wait for that until approving this. I think these 2 features should be strictly tied together. If somebody would use these current changes only, it would be quite sure that they would get the same mail mutlipel times.

Good job so far!

karisal-anders:
Not needed, needed functionality was implemented in PR #384. Closing as unneeded.

PR #385, opened 2024-06-14
KK-1184 | feat: batch user is_obsolete updating in auth change command

karisal-anders:
### feat: batch user is_obsolete updating in auth change command

refs KK-1184

## Related

[KK-1184](https://helsinkisolutionoffice.atlassian.net/browse/KK-1184)

[KK-1184]: https://helsinkisolutionoffice.atlassian.net/browse/KK-1184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/users/management/commands/send_user_auth_service_is_changing_notifications.py
+++ b/users/management/commands/send_user_auth_service_is_changing_notifications.py
@@ -43,7 +43,7 @@ class Command(BaseCommand):
             "-b",
             "--batch_size",
             type=int,
-            help="Batch size for bulk updates and queryset iteration. "
+            help="Batch size for queryset iteration and data updating. "
             "By default %(default)s.",
             default=1000,
         )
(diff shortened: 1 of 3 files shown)


nikomakela (commented):
https://www.reddit.com/r/django/s/mM375ppOaf

This says the update is the best when all the rows are updated to the same value. I'm not sure how does the amount of rows matter in the generated sql query. My assumption is that update-method uses update where id in -query. What happens when there are 20k ids in the list?

The same article however says that the bulk_update would need to fetch all the rows first. And it should use the atomic process too?

In our case, to me it looks like the update would be better option.

nikomakela:
There seems to be no limit in Postgres in amount of arguments https://stackoverflow.com/questions/1009706/postgresql-max-number-of-parameters-in-in-clause

nikomakela (approved):

PR #426, opened 2024-10-15
feat: project message permissions for sending to all

nikomakela:
KK-1039 KK-1243.

Add a project permission that defines whether a user can send a message
to all members in a project. The new project permission will be
available to be selected in project object permissions, and in user and
group admin forms.

Also refactored the other existing project permissions so that they are
all listed in a new enum. The new enum helps the future refactoring of
the code, because it is now a referable constant.

----

fix(admin): performance and UX issue in messaging admin


terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr426.api.dev.hel.ninja :laughing::tada::tada::tada:

messaging/schema.py
```diff
@@ -277,11 +292,13 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         if recipient_selection := data.get("recipient_selection"):
             validate_recipient_selection_and_data(
                 recipient_selection,
-                data["event"] if "event" in data else message.event,
+                data.get("event", message.event),
```
karisal-anders, line 295: This is probably ok, just saying that now message.event is evaluated (Not returned, but evaluated) always regardless of whether `"event" in data` is true or not. The possible problem could be if evaluating `message.event` would cause a problem, but it's a ForeignKey in Message so probably it's ok, I think it'll return None if the ForeignKey is not set (blank=True at least in the model)—didn't check this though.

messaging/schema.py
```diff
@@ -327,7 +359,15 @@ class Input:
...
-
+        # user = info.context.user
+
+        # if validate_recipient_selection_and_data(
+        #     message.recipient_selection,
+        #     message.event,
+        #     message.occurrences,
+        #     user,
+        #     message.project,
+        # ):
```
karisal-anders, line 370: Why is there commented out code here?

common/tests/conftest.py
```diff
@@ -108,27 +111,27 @@ def guardian_api_client():
...
     user = UserFactory()
-    assign_perm("admin", user, project)
+    assign_perm(ProjectPermission.ADMIN.value, user, project)
 
     if request.param:
-        assign_perm("publish", user, project)
+        assign_perm(ProjectPermission.PUBLISH.value, user, project)
     else:
-        assign_perm("projects.publish", user)
+        assign_perm(f"projects.{ProjectPermission.PUBLISH.value}", user)
```
karisal-anders, line 126: Would it be better to import `PERM_CAN_PUBLISH_EVENTS` from projects.models and use it here?
> PERM_CAN_PUBLISH_EVENTS = f"projects.{ProjectPermission.PUBLISH.value}"

karisal-anders (approved):
Left a few minor comments, please take a look at them. The commented out code should be removed IMHO unless there's a good reason for including it (And in that case the reason would be good to include as a comment in the code IMHO).

messaging/schema.py
nikomakela: Should have had it as a draft or something. There is now an actual permission check there. I was jsut waiting that I'll test the feature with admin UI first.

nikomakela:
Added [feat: add canSendToAllInProject permission to myAdminProfile query](https://github.com/City-of-Helsinki/kukkuu/pull/426/commits/6916c5d25e81f8646ab6aa95ba1faea76e609150)

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr426.api.dev.hel.ninja :laughing::tada::tada::tada:

PR #435, opened 2024-11-28
Test ruff is same in requirements-dev.txt and .pre-commit-config.yaml

karisal-anders:
### test: ruff is same in requirements-dev.txt and .pre-commit-config.yaml

also:
 - update ruff to same version in .pre-commit-config.yaml as it is in
   requirements-dev.txt, they were out of sync

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,8 +8,11 @@ default_language_version:
   python: python3.11
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version
-    rev: v0.5.2
+    # NOTE: Don't move the trailing "ruff-pre-commit version" comment from
+    # the ruff's version line, it is used by test_pre_commit_ruff_version.py
+    # test case in order not to have to add a YAML library dependency just
+    # to test this version:
+    rev: v0.8.0  # ruff-pre-commit version
     hooks:
       # Run the linter
       - id: ruff
--- a/kukkuu/tests/test_pre_commit_ruff_version.py
+++ b/kukkuu/tests/test_pre_commit_ruff_version.py
@@ -1,0 +1,40 @@
+import re
+
+RUFF_VERSION_PRE_COMMIT_REGEX = re.compile(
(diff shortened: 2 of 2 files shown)


terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr435.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://kukkuu-pr435.api.dev.hel.ninja :laughing::tada::tada::tada:

nikomakela (approved):
Using the pytest for this might be an overkill. The pytests are now also a bit fragile, because they rely parsing text from many files. Small changes to the formatting of these files could break the tests, even if the Ruff versions are still consistent. 

Instead of directly testing the functionality (consistent linting), these tests focus on an implementation detail (matching versions).

I would say that it is quite clear that the Ruff version is not the same with the pyenv Ruff, if the precommit hook gives a different result as the other Ruff gave. If the result is the same, it does not even matter if the versions are not matching. 

I think that the best solution would be if we could read the version from `pyproject.toml` also for the precommit hook configuration, but that might be too problematic.

While we still don't have anything else to offer, we can go with this, but I would be OK even without the pytests. Good that the version is now upgraded for precommit hook.

karisal-anders:
> Using the pytest for this might be an overkill. The pytests are now also a bit fragile, because they rely parsing text from many files. Small changes to the formatting of these files could break the tests, even if the Ruff versions are still consistent.

Yes, they're a bit fragile, but they're named in a way that they tell what is broken if they break, and they should be easy to fix if they do break.

> Instead of directly testing the functionality (consistent linting), these tests focus on an implementation detail (matching versions).

I don't see any reason to allow different versions of ruff being used for different parts of the same system. It just doesn't make any sense to me that we would be using different version of ruff in the docker container and then another in precommit hook. So checking or enforcing that they're the same version makes sense to me.

> I think that the best solution would be if we could read the version from `pyproject.toml` also for the precommit hook configuration, but that might be too problematic.

That would be a better solution IMHO too, if that'd be possible.

PR #454, opened 2025-01-03
fix(auditlog-extra): admin mixin should use get_changelist_instance

nikomakela:
KK-1113.

In some cases, the context does not have the changelist initialized, so it is better to fetch the change list instance with `ModelAdmin.get_changelist_instance`. Also, the post requests which are only used by the django admin actions are just logging the reads as duplicated.


karisal-anders (approved):
The now unnecessary comment in changelist_view's docstring could be removed but otherwise looks good:
```
        The changelist in the response's context_data (i.e. "cl") should be set by the
        super().changelist_view call, see e.g.
        https://github.com/django/django/blob/4.2.16/django/contrib/admin/options.py#L2071
```

PR #502, opened 2025-05-23
chore: migrate all obsoleted users to be deactivated

nikomakela:
KK-1251.

A new Django data migration that sets all the obsoleted users also as deactivated.


karisal-anders (approved):
Looks ok to me, but I assume that Ratkaisutoimisto would want to code review this on their own before merging is possible.

PR #501, opened 2025-05-23
feat: send notification only to active and non-oboleted users

nikomakela:
KK-1251

NOTE: check also https://github.com/City-of-Helsinki/kukkuu/pull/502 for data migration and https://github.com/City-of-Helsinki/kukkuu/pull/503 to fully remove the `is_obsolete` field from User model.

PR #503, opened 2025-05-23
feat: remove is_obsolete field from user model

nikomakela:
KK-1251.

The `is_obsolete` field in User model was only used during the migration process when authentication service was changed from Tunnistamo to a Keycloak service of Helsinki Profile. It was deprecated after all the users were (data) migrated so that all the obsoleted users were also deactivated. There is no need for user.is_obsolete anymore, so it could be deleted.


nikomakela:
Merged to https://github.com/City-of-Helsinki/kukkuu/pull/501 via https://github.com/City-of-Helsinki/kukkuu/pull/502

PR #506, opened 2025-06-02
CSV export view for NotificationTemplates and support for file imports

nikomakela:
KK-1441.

Replace the old Importers app with a clone from Kultus (https://github.com/City-of-Helsinki/palvelutarjotin/tree/80e9f26097825fee03012e2a74e02e0398eeb705/notification_importers). The improved version of notification importers is first copied with minimal changes in order to get it to work in Kukkuu. Then it's extended with an export to CSV feature. The long goal is to get rid of Google Drive Spreadsheet that stores the content for notification templates. This new Django app looks the same as the old one, but also supports importing the notification templates from the code repo itself. It's been in use in Kultus for a long now.

image

NOTE: This PR also contains small fixes to CSP rules.

PR #509, opened 2025-06-05
Add notification templates in code repository in HTML format

nikomakela:
KK-1441.

Based on https://github.com/City-of-Helsinki/kukkuu/pull/506, which starts using the `notification_importer` app (that was copied from [Kultus](https://github.com/City-of-Helsinki/palvelutarjotin/tree/d211e7fc841188194d2a03896c1f127b4720e28c/notification_importers)), but with a Google Spreadsheet importer.

In this PR, the notification templates content is copied from https://docs.google.com/spreadsheets/d/1TkdQsO50DHOg5pi1JhzudOL1GKpiK-V2DCIoAipKj-M. The content is then added to notification template HTML files that can be imported with `notification_importers.notification_importer.NotificationFileImporter`.

This changes the Google Spreadsheet importer to a file importer, so the notification templates can be fetched from the project's repository. The templates needs to be under `notification_importers` app in templates -folder. They should be categorized by the notification type, so the emai ltemplates should be under notification_importers/templates/email/ -folder. Then the files should be named so that the name is constracted from `[type]-[locale].html`.

The notification importer needs `<title>` -tag in order to set subject field properly. Also the HTML syntax makes the template look better when sent to customer. However, since there is no design for those templates yet, the HTML content should be as simple as possible (e.g. similar to markdown format). Target was to convert plain text messages to minimal HTML equivalent.

Instead of showing the long and ugly URL, now when we are using HTML-formatted notification templates, we could also use some text as a link instead of the URL (as text).

image

### DIFF
--- a/kukkuu/settings.py
+++ b/kukkuu/settings.py
@@ -474,8 +474,12 @@ KUKKUU_FEEDBACK_NOTIFICATION_DELAY = env("KUKKUU_FEEDBACK_NOTIFICATION_DELAY")
 # If set to 0, unenrolment is allowed till the occurrence start.
 KUKKUU_ENROLMENT_UNENROL_HOURS_BEFORE = env("KUKKUU_ENROLMENT_UNENROL_HOURS_BEFORE")
 NOTIFICATIONS_IMPORTER = (
-    "notification_importers.notification_importer.NotificationGoogleSheetImporter"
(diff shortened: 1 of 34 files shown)


nikomakela:
Converted to draft, since there is no plan for this yet and Solution office requested the change to draft.

PR #517, opened 2025-07-07
KK-1462: Update Azure storage settings

mhieta:
Update Azure storage settings.

Refs: KK-1462


nikomakela (approved):
It seems like if "SharedAccessSignature" is in connection string, it's always visible. Another way to configure it should be AZURE_SAS_TOKEN. Let's see how it works in dev env.

terovirtanen:
**KUKKUU-API branch is deployed to platta: https://kukkuu-pr517.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #522, opened 2025-08-21
feat: add resilient_logger for audit log transport

JokinenK:


tuomas777 (approved):
The changes seem to contain some minor formatting issues, which should be fixed with `pre-commit`. Other than that, LGTM 👍

PR #601, opened 2026-05-25
fix(audit_log): exclude last_api_use from audit logging

voneiden:
Refs: KK-1535


charn (approved):

kukkuu/auditlog_settings.py
```diff
@@ -54,6 +54,7 @@
         "model": "users.user",
         "exclude_fields": [
             "last_login",  # don't write log of every request
+            "last_api_use",  # helusers internal, changes on every API request
```
charn: This is a date field so, it changes only once per day.

voneiden: oh true!

